PR #103, opened 2023-09-11
Add support for labelled items, and VCA

### LINKED ISSUE #102: Custom site labels
CASTEP supports custom site labels, specified in the .cell file like this:

```
%BLOCK positions_abs
H  0.0  0.0  0.0 LABEL=H1
H  0.0  0.0  1.0 LABEL=H2
%ENDBLOCK positions_abs

```

(Note that this is different from the `H:H1  0.0  0.0  0.0 ` syntax that would be used to specify custom species. )

Near the start of the .castep file you get a block that looks like this:
```
     xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
     x  Element    Atom        Fractional coordinates of atoms  User-defined  x
     x            Number           u          v          w          LABEL     x
     x------------------------------------------------------------------------x
     x  H            1         0.000000   0.000000   0.000000     H1          x 
     x  H            2        -0.000000  -0.000000   0.166667     H2          x 
     xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx



```

Crucially, for atom site properties, e.g. forces, you get this sort of thing:

```
 ************************************** Forces **************************************
 *                                                                                  *
 *                           Cartesian components (eV/A)                            *
 * -------------------------------------------------------------------------------- *
 *                         x                    y                    z              *
 *                                                                                  *
 * H [H1]          1      0.00000              0.00000             -0.06470         *
 * H [H2]          2     -0.00000             -0.00000              0.06470         *
 *                                                                                  *
 ************************************************************************************
```

I got a bit lost in the regex trying to figure out where to change the code to accommodate such custom labels.

oerc0122:
Fixes #102

### DIFF
--- a/castep_outputs/castep_file_parser.py
+++ b/castep_outputs/castep_file_parser.py
@@ -482,6 +482,45 @@ def parse_castep_file(castep_file_in: TextIO,
             curr_run[mode] = _process_atreg_block(block)
 
         # Initial pos
+        elif block := get_block(line, castep_file,  # Labelled
+                                r"Fractional coordinates of atoms\s+User-defined",
+                                gen_table_re("", "x+")):
+
+            if "position" not in to_parse:
+                continue
+
+            logger("Found initial positions")
+
+            curr_run["initial_positions"] = {}
+            for line in block:
+                if match := REs.LABELLED_POS_RE.search(line):
+                    ind = f"{match['spec'].strip()} [{match['label'].strip()}]", int(match["index"])
+                    curr_run["initial_positions"][ind] = to_type(match.group("x", "y", "z"), float)
+
+        elif block := get_block(line, castep_file,  # Mixture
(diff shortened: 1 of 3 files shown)


jkshenton (approved):
Thanks! 

I've tested this and it now works for both the species tagging and the custom site labels. I haven't checked the VCA, but it looks good to me. 

The resulting keys for the site properties get a little unwieldy with the tags and the labels - I think we should either reorganise the way these are stored/ provide some sort of aliases/ have utility functions to extract the most common bits of information needed. But that can be a in separate PR.

PR #119, opened 2024-01-31
Make filters a `Flag` Enum to allow combinations

oerc0122:


castep_outputs/castep_file_parser.py
```diff
@@ -31,40 +32,67 @@
...
+              "HIGH": HIGH,
+              "TESTING": TESTING,
+              "FULL": Filters,
               }
 
 
 def parse_castep_file(castep_file_in: TextIO,
-                      filters: Optional[Tuple[Literal[DATA_CLASSES]]] = None,
+                      filters: Optional[Filters] = None,
                       level: Literal[DATA_LEVEL.keys()] = "FULL") -> List[Dict[str, Any]]:
```
ajjackson, line 96: It looks like if `filters` and `level` are both set, `level` will be ignored. Is that likely to cause surprises?

castep_outputs/castep_file_parser.py
```diff
@@ -31,40 +32,67 @@
...
+
+
+LOW = (Filters.BS | Filters.CELL | Filters.CHEM_SHIELDING | Filters.DIPOLE |
+       Filters.ELASTIC | Filters.ELF | Filters.FINAL_CONFIG | Filters.FORCE |
+       Filters.MD_SUMMARY | Filters.OPTICS | Filters.POPN_ANALYSIS |
+       Filters.POSITION | Filters.SOLVATION | Filters.SPECIES_PROPS |
+       Filters.STRESS | Filters.TDDFT | Filters.THERMODYNAMICS | Filters.TSS)
+
+MEDIUM = LOW | Filters.PARAMETERS | Filters.GEOM_OPT | Filters.MD | Filters.PHONON
+HIGH = MEDIUM | Filters.PSPOT | Filters.SCF | Filters.SYMMETRIES | Filters.SYS_INFO
```
ajjackson, line 76: Very pretty!

ajjackson (commented):
This seems very nice. The enums improve clarity as we see that we are comparing flags for high-level logic; matching strings feels more like parsing and so is confusible with lower-level operations.

castep_outputs/castep_file_parser.py
oerc0122: Good point. I've made it so that:
- `level`and `filters` (now `extra_filters`) must be passed as kwargs
- If neither passed, `level` defaults to `HIGH`
- If `extra_filters` passed `level` defaults to `NONE`, it is then subsequently `or`'d with `extra_filters`

ajjackson: That is a very useful combination of defaults, but maybe a little complicated to explain/remember/debug? Who is the end-user for this function?

An alternative scheme for consideration:
- filters defaults to `None`
- level defaults to `None`
- They are always `or`ed
- If both are `None`, nothing is parsed

oerc0122: The reason I do this is that if you are a general user you probably just want to just call `parse_castep_file` and trust the output only advanced users are going to want to fiddle with this.

oerc0122: The other option is that I remove `level` entirely, and if people want to set the filters manually, they import the module variables, which could be attached to the `Filters` `Enum`, this may be a better approach

ajjackson: Then maybe
```python
def parse_castep_file(castep_file_in: TextIO,
                      level: Literal[DATA_LEVEL.keys()] = "HIGH",
                      extra_filters: Optional[Filters] = None) -> List[Dict[str, Any]]:
```
and we trust advanced users to set level "NONE" as appropriate? "extra_filters" makes it clear they operate on top of the level.

ajjackson: I bring this up to atone for similarly helpful-but-complicated default arguments I have implemented in the past 😉

ajjackson: > The other option is that I remove `level` entirely, and if people want to set the filters manually, they import the module variables, which could be attached to the `Filters` `Enum`, this may be a better approach

It is a lot simpler, which is promising! And then we can still have the "HIGH" collection as a default, which also provides a hint of where to look for the others.

ajjackson: Yeah, this looks great. And allows one to do things like `Filters.HIGH & ~Filters.MEDIUM` 🙃

ajjackson:
I think this needs a test case with non-trivial filtering.

oerc0122:
Right, they're pretty rough tests, but there's still a lot of solidifying work that needs to happen with regards to testing. More datasets with better representation.

castep_outputs/test/test_filters.py
```diff
@@ -0,0 +1,46 @@
...
+
+class test_castep_filters(TestCase):
+    def setUp(self):
+        self.test_text = _TEST_FILE.open("r", encoding="utf-8")
+
+    def tearDown(self):
+        self.test_text.close()
+
+    def test_filter_all(self):
+        for filtery, keys in ((Filters.CELL, ('initial_cell',)),
```
ajjackson, line 21: Perhaps it would be a good idea to use `subTest` here to help make sense of test failures:

https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

castep_outputs/test/test_castep_parser.py
```diff
@@ -1595,19 +1595,6 @@ def test_get_md_data(self):
                            ('Si', 2): (0.000729, 0.504323, 0.520521)},
              'memory_estimate': [{'model_and_support_data': {'disk': 0.0, 'memory': 27.2},
                                   'molecular_dynamics_requirements': {'disk': 0.0, 'memory': 13.4}}],
-             'scf': [[{'energy': -854.989696,
-                       'energy_gain': None,
-                       'fermi_energy': 0.0,
-                       'time': 17.06},
-                      {'energy': -855.716038,
-                       'energy_gain': 0.0907927915,
```
ajjackson, line 1603: Is this section removed because of a change to the default filter level?

castep_outputs/test/test_filters.py
oerc0122: Knew there was a trick for this, thanks, I had forgotten `subTest` despite using it before.

ajjackson: I happened to read about it recently! Haven't tried it out yet (most projects moved to pytest anyway...) but seems a good fit here.

ajjackson (approved):
I think that's everything

castep_outputs/test/test_filters.py
oerc0122: Yeah, `pytest` and friends add a lot of useful stuff (looking at you `parametrize`), but stdlib-only is such a boon here for portability given its intended use.

PR #123, opened 2024-02-16
Cleanup SCF debug info

oerc0122:
Move extra SCF information (max verbose) into its own sub-`dict` due to potential overlap with main SCF info (`fermi energy` vs `fermi_energy`) if all dictionary keys eventually normalised.


ajjackson:
Are the "debug information" Fermi energy and the calculated/reported Fermi energy two different pieces of information? I see they are reported to different precision in the test data.

I also see that the test data has debug information for one SCF step but not the next, is that typical?

How nicely does this interact with finite-basis corrections, which add on a few mini-SCFs for each "main" singlepoint SCF? Are they another dict item?

oerc0122:
The debug fermi and the final fermi should be the same (but to different precision), but sometimes I think it might make multiple iterations and potentially backtrack in the solver in more complex cases. 

You're right, the debug info is attached to the wrong cycle. 

Not sure on the mini-scfs, at the moment, this should be general for most SCFs (though it really depends how much we expect someone to want to extract the debug info here, it's only available on `verbosity` >3 i.e. Phil. This is reasonably low priority, hence the separation from main SCF info)

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -1622,14 +1627,16 @@ def _process_scf(block: Block) -> list[SCFReport]:
...
+
         elif "no. bands" in line:
-            curr["no_bands"] = to_type(get_numbers(line)[0], int)
+            debug_info["no_bands"] = to_type(get_numbers(line)[0], int)
 
         elif "Kinetic eigenvalue" in line:
-            if "kinetic_eigenvalue" not in curr:
-                curr["eigenvalue"] = []
+            debug_info.setdefault("kinetic_eigenvalue", [])
+            debug_info.setdefault("eigenvalue", [])
```
ajjackson, line 1637: Default for "eigenvalue" is set up here, but unlike "kinetic_eigenvalue" the values are appended a bit further down in a different `elif`. Is that ok, is this block guaranteed to come first?

ajjackson (approved):
Looks good! One quick query

castep_outputs/parsers/castep_file_parser.py
oerc0122: Yeah it's guaranteed to come first and note that it's `setdefault` so if it occur again it's not reset.

PR #125, opened 2024-02-17
Normalise dict keys for easier processing

oerc0122:
Make most `dict` keys underscore-separated lower-case strings


ajjackson (approved):
This looks healthy to me: it reduces the number of bits of information a developer needs to remember, without severely impacting readability. The main acronyms that suffer are BLAS and FFT, and these are so frequently encountered that I didn't really stumble on the lower-case versions.

Is it worth compiling the regexp, as it gets used a lot? Feel free to declare that "premature optimisation" and ignore.

PR #127, opened 2024-02-21
Fix issue with MD missing step 0 and capturing memory estimate

oerc0122:
- Add new capture block for step 0 (**N.B.** does not have `Iteration` in line)
- If `memory_est` in MD step, push it up to top level.


castep_outputs/parsers/castep_file_parser.py
```diff
@@ -966,7 +966,30 @@ def parse_castep_file(castep_file_in: TextIO,
...
+            # Put memory estimate to top level
+            # TODO: Check if memory estimate can appear multiple times
+            if "memory_estimate" in data:
+                curr_run["memory_estimate"] = data.pop("memory_estimate")
+
+            curr_run["md"].append(data)
+
+        elif block := Block.from_re(line, castep_file,  # Capture 0th iteration
+                                    "Starting MD",
+                                    gen_table_re("", "=+")):
```
ajjackson, line 987: Is there any difference in handling between the "Starting MD" block and "Starting MD iteration" block? At a glance the code looks identical except for the `if "memory_estimate"` block and its not obvious that this could not safely see both block types.

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -966,7 +966,30 @@ def parse_castep_file(castep_file_in: TextIO,
...
+
+            # Avoid infinite recursion
+            next(block)
+            data = parse_castep_file(block)[0]
+            add_aliases(data, {"initial_positions": "positions",
+                               "initial_cell": "cell"},
+                        replace=True)
+
+            # Put memory estimate to top level
+            # TODO: Check if memory estimate can appear multiple times
```
ajjackson, line 979: Is there a plan/Issue to check this? I think @DBJochym worked on the Mac memory tracking at one point but maybe that is disconnected from these estimates.

ajjackson (commented):
Looks pretty sensible, just a couple of comments

castep_outputs/parsers/castep_file_parser.py
oerc0122: This is the initial memory estimate based on predicted allocation rather than actual memory measurement. In the case of MD it appears after the `starting MD` block, which is why it needs to be pushed up. The problem is I don't know if higher verbosities repeat the info.

oerc0122: As for the issue it's partly covered by #1

ajjackson: Ok! I'm not a big fan of TODO in the codebase, either:
- there should be an automated procedure to catch and review TODO comments; or
- the TODO should be removed and made into an explicit issue. (Could just be a bullet point in #1).

Otherwise they tend to multiply and live forever.

oerc0122:
```suggestion
```

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -967,7 +967,30 @@ def parse_castep_file(castep_file_in: TextIO,
             if Filters.MD not in to_parse:
                 continue
 
-            logger("Found MD Block (step %d)", len(curr_run["md"])+1)
+            logger("Found MD Block (step %d)", len(curr_run["md"]))
+
+            # Avoid infinite recursion
+            next(block)
+            data = parse_castep_file(block)[0]
```
ajjackson, line 974: How about

```python
data, *rest = parse_castep_file(block)
assert rest == []
```

oerc0122: Hm. The alternative would be:
```suggestion
            data, = parse_castep_file(block)
```
Which would also do that (though the comma being significant is not obvious and `ruff` would demand:
```suggestion
            (data,) = parse_castep_file(block)
```
Not sure I actually like the look of any of them or think they add clarity. `parse_castep_file` should, in principle, handle any unexpected structures which might result in multiple elements in the list (and error if the calculation ended in a partial state).

castep_outputs/parsers/castep_file_parser.py
oerc0122: Reflecting on this, it should be easy to add the alternation. Not sure what past me was on about. I'll take a look when I get a moment.

ajjackson (approved):

PR #142, opened 2024-07-28
Refactor magres parsing

### LINKED ISSUE #139: Rework magres parser
Magres parser doesn't handle `old_magres` section and `calc` properties should move to separate dict.

oerc0122:
Fixes #139


castep_outputs/parsers/magres_file_parser.py
```diff
@@ -21,40 +40,138 @@ def parse_magres_file(magres_file: TextIO) -> Dict[str, float]:
...
+    dict[str, str]
+        Processed data.
+    """
+    accum = {"units": {}, "ms": {}}
+    for line in block:
+        if line.startswith("ms"):
 
-                        key, spec, ind, *val = blk_line.split()
-                        accum[key][(spec, int(ind))] = to_type(val, float)
+            key, spec, ind, *val = line.split()
```
jkshenton, line 131: There's an issue with some legacy magres files where if there is more than 99 atoms of a given type, the index runs into the spec label for **ms** and **efg** lines. One solution here could be to simply count the number of elements on this line (split by whitespace). It should be 3 + 3x3 = 12. If the length is 11, then use regex to split the spec-ind. 

For example, changing this line to e.g. 
```python
# Regex pattern to separate label and index
pattern = re.compile(r"([a-zA-Z]+[0-9]*[a-zA-Z]*)([0-9]+)")

line_list = line.split()
if len(line_list) == 12:
    key, spec, ind, *val = line_list
elif len(line_list) == 11:
    key, specind, *val = line_list
    # Apply regex to separate specind into spec and ind
    match = pattern.match(specind)
    if match:
        spec = match.group(1)
        ind = match.group(2)
    else:
    # some warning?
        print(f"No match for line: {line_list}")
    
```

However, people also use custom site labels (from cif files) which can have numbers in them - e.g. H1, H2a - can be anything really... so there's a chance that blindly using regex to split could fail in some cases. However, I think these would be quite rare compared to the more common spec-index run-in. A more complex solution would perhaps need to keep track of the labels from the atoms block and then use them in the parsing of the ms and efg lines.

oerc0122: Yeah, that's the trouble. I've put in an awful hack which verifies the version of castep > 22 it checks the last 5 if it thinks values are munged together.

ajjackson: Alteratively: when this ambiguity is detected you raise an error "cannot read Magres file of > 99 atoms from CASTEP < 23 due to ambiguous formatting. Please insert spaces or use a newer version of CASTEP".

jkshenton: Thanks @oerc0122 ! 

I think this will probably cover the majority of the issues in legacy files so I think a _warning_ (asking the user to double-check the labels/indices) would be enough. I get questions about this problem sufficiently often to suggest that it's worth having a hack like this to cover most cases. 

CASTEP 22.1.1 has the 'fix' so should be included - I think the check is correct in the https://github.com/oerc0122/castep_outputs/pull/142/commits/76426b91c3ec15a63d0437cb485e5e427deb3654 commit, but just noting it here.

oerc0122: The reason I wrote it this way was that I started implemented the over-complicated logic required for the 2-specind component of the j-coupling lines. Only once I'd solved the logic and implemented it did I notice in the castep source that the `isc` prefixed vars all have enforced spaces.

ajjackson (approved):
LGTM

PR #144, opened 2024-07-28
Fix parsing of bands file header

### LINKED ISSUE #137: Bands file header not picked up (not standard)

oerc0122:
Fixes #137

### DIFF
--- a/castep_outputs/test/bands.yaml
+++ b/castep_outputs/test/bands.yaml
@@ -1,3 +1,4 @@
+Fermi energy: [0.323077]
 bands:
 - band: [-2.56339233, -2.56319008, -2.5626493]
   qpt: [-0.375, -0.375, -0.375]
@@ -7,3 +8,12 @@ bands:
   qpt: [-0.375, -0.375, -0.125]
   spin_comp: 1
   weight: 0.03125
+coords: {}
+eigenvalues: 3
+electrons: 300
+k-points: 2
+spin components: 1
+unit_cell:
+- [11.932676, 0.0, 0.0]
+- [0.0, 11.932676, 0.0]
+- [0.0, 0.0, 11.932676]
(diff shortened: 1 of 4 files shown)


castep_outputs/test/bands.yaml
```diff
@@ -1,3 +1,4 @@
+Fermi energy: [0.323077]
```
ajjackson, line 1: Why is Fermi energy wrapped in a list, when other scalar attributes appear on their own?

ajjackson (commented):
The code looks fine to me, but I don't entirely follow what bug is being fixed. Am I right in thinking that it was looking for (nonexistent) "BEGIN header" and "END header" fields? We can confirm from the docs that this is not expected, so this seems good https://www.tcm.phy.cam.ac.uk/castep/documentation/WebHelp/content/modules/castep/expcastepfilebands.htm

One query about scalar data

oerc0122:
> The code looks fine to me, but I don't entirely follow what bug is being fixed. Am I right in thinking that it was looking for (nonexistent) "BEGIN header" and "END header" fields? We can confirm from the docs that this is not expected, so this seems good https://www.tcm.phy.cam.ac.uk/castep/documentation/WebHelp/content/modules/castep/expcastepfilebands.htm
> 
> One query about scalar data

That's correct. I had assumed at first glance that it was a "standard header". Sadly as all things scientific software, standards are there to be broken.

castep_outputs/test/bands.yaml
oerc0122: Another change that needs to happen is normalising the keys after they're read. I suspect I'll fix parse standard header in another PR.

ajjackson: > It's a bit of a hack I may clean up later with a if len(data) == 1: data[0] style thing.

Not sure I like the sound of that, surely there are some variable-length cases which this would make unsafe to work with.

ajjackson (approved):
Happy that this PR is achieving its goal and tested appropriately

PR #145, opened 2024-07-28
Fix per-species capture in phonon dos

### LINKED ISSUE #138: phonon_dos not properly extracting species DOS

oerc0122:
Fixes #138


castep_outputs/parsers/phonon_dos_file_parser.py
```diff
@@ -57,14 +57,17 @@ def fix(qdat):
             logger("Found DOS block")
 
             dos = defaultdict(list)
+            # First chunk is " BEGIN DOS   Freq (cm-1)  g(f)", thus need the 5th on
+            species = map(str.lower, block[0].split()[5:])
```
ajjackson, line 61: Probably you already thought about this, but it's weird enough to ask about:

Is it _really_ appropriate to downcase element names and user labels? Can that ever lead to ambiguity? (I _think_ we're OK on the ambiguity front because of the colon. I hope CASTEP doesn't allow multiple user labels distinguished by case. I hope.)

ajjackson (dismissed):
Code looks nice and test coverage is 100%

The test case only covers a single species; I checked against the quartz .phonon_dos from Euphonic test data and the resulting DOS dict has "g", "o" and "si" sections as expected. (Capitalisation aside...)

castep_outputs/parsers/phonon_dos_file_parser.py
oerc0122: That is a good point, I think it was leftover from when I was trying to parse all the data-sets in one block.
```suggestion
            species = block[0].split()[5:]
```

ajjackson (approved):
LGTM

Issue #10, opened 2023-07-21
Add sphinx documentation


oerc0122:

oerc0122:
Basic documentation since #51

PR #146, opened 2024-07-28
Docstrings

### LINKED ISSUE #132: Rename AP in `cli.args`
ALLCAPS name not great for an object instance. Out-of-scope for this PR though!

_Originally posted by @ajjackson in https://github.com/oerc0122/castep_outputs/pull/116#discussion_r1679455366_

oerc0122:
Extensive rework of documentation (still ongoing):
- Adds numpy docstrings to all functions.
- Adds TypedDict return docs to all file-parsers (except complex ones such as cell/param/castep).
- Makes several file-parsers slightly more consistent in handling of params.
- Add docstring tests to relevant utility functions.

Fixes #132 
Address #10


castep_outputs/utilities/castep_res.py
```diff
@@ -8,20 +8,72 @@
...
+
+    Parameters
+    ----------
+    labels : ~collections.abc.Sequence[str]
+        Iterable of labels to label each group.
+    counts : ~collections.abc.Sequence[int | str | None]
+        Iterable of counts to group into each label (count must not exceed that of labels).
+    sep : str
+        Separator between floats.
+    suff : str
```
ajjackson, line 54: I find this abbreviation hard to parse at speed, is it worth it for two characters?

castep_outputs/utilities/castep_res.py
```diff
@@ -8,20 +8,72 @@
...
+
+    Returns
+    -------
+    str
+        Regular expression to match floats with given labels.
+
+    Raises
+    ------
+    NotImplementedError
+        If `suffix` and `counts` provided.
```
ajjackson, line 65: "Suffix" here is not consistent with (but preferable to) the function signature

castep_outputs/utilities/castep_res.py
```diff
@@ -38,8 +90,42 @@ def labelled_floats(labels: Sequence[str], counts: Sequence[int | str | None] =
 
 def gen_table_re(content: str, border: str = r"\s*",
                  *, pre: str = "", post: str = "", whole_line: bool = True):
```
ajjackson, line 92: Can we get a return type here?

castep_outputs/utilities/castep_res.py
```diff
@@ -53,43 +139,113 @@ def gen_table_re(content: str, border: str = r"\s*",
...
             if val}
 
 
 # --- RegExes
 # Regexps to recognise numbers
+#: Floating point number.
+#:
+#: :meta hide-value:
 FNUMBER_RE = r"(?:[+-]?(?:\d*\.\d+|\d+\.\d*))"
+#: Integer number.
```
ajjackson, line 177: I'm finding it confusing that there is an empty-looking between this "title" and the value it refers to... but not after the value. So at a glance it looks like FNUMBER_RE is an integer, EXPFNUMBER_RE is "integer ratio" and so on.

castep_outputs/utilities/datatypes.py
```diff
@@ -181,154 +332,354 @@ class MullikenInfo(TypedDict, total=False):
...
+    #:
+    #: - ``U`` - a single ultrasoft projector.
+    #: - ``UU`` - Two ultrasoft projectors.
+    #: - ``N`` - a single norm-conserving projector.
+    #: - ``L`` - use this projector as the local component.
+    #: - ``G`` - an ultrasoft GIPAW Gamma projector.
+    #: - ``H`` - an norm-conserving GIPAW Gamma projector.
+    #: - ``P`` - Dummy: do not make a projector.
+    #: - ``LG`` - Make Gammas for local channel (not done by default).
+    type: Literal["U", "UU", "N", "L", "G", "H", "P", "LG", "LL", "GG", "LGG", None]
```
ajjackson, line 353: I see `None` is an undocumented but allowed option. Does this represent optional/missing information? Is it correct that this is optional but the other keys are not?

castep_outputs/utilities/datatypes.py
```diff
@@ -181,154 +332,354 @@ class MullikenInfo(TypedDict, total=False):
...
     fine: float
+    #: Projection
     proj: str
+    #: All pseudopotential projectors.
     projectors: tuple[PSPotProj, ...]
+    #: Extra options.
     opt: list[str]
+    #: Swap particular orbitals.
     shell_swp: str
+    #: ?
```
ajjackson, line 382: ?

castep_outputs/utilities/dumpers.py
```diff
@@ -17,38 +18,106 @@
...
+        Dumping function.
+
+    Raises
+    ------
+    ValueError
+        Invalid `dump_fmt` provided.
+    ImportError
+        No valid YAML dumper and `yaml` requested.
     """
     if dump_fmt not in SUPPORTED_FORMATS:
```
ajjackson, line 118: Should the `dump_fmt` be typed as a literal? Could be annoying to maintain as the ground truth is a dict.

castep_outputs/utilities/utility.py
```diff
@@ -114,8 +247,17 @@ def flatten_dict(dictionary: MutableMapping[Any, Any], *,
     return dict(items)
 
 
-def stack_dict(out_dict: dict[Any, list], in_dict: dict[Any, list]) -> dict[Any, list]:
-    """ Append items in `in_dict` to the keys in `out_dict` """
+def stack_dict(out_dict: dict[Any, list], in_dict: dict[Any, list]):
```
ajjackson, line 250: An explicit `) -> None` would be nice on this one.

ajjackson (commented):
Lots of work here! A few comments to look at

ajjackson:
This is a big boost to readability of the codebase which makes it much easier to reason about what functions will do and what return values to expect 🚀

castep_outputs/parsers/__init__.py
```diff
@@ -48,6 +48,7 @@
            "parse_phonon_file"]
 
 
+#: Dictionary of available parsers.
```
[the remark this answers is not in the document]
oerc0122: This *is* a docstring for the constant. Hence why I've added it.

castep_outputs/cli/castep_outputs_main.py
```diff
@@ -60,9 +86,30 @@ def parse_single(in_file: str | Path | TextIO,
...
+        loglevel: int = logging.WARNING,
+        testing: bool = False,
+        **files,
+):
+    """
+    Parse all files in files dict.
+
+    Parameters
+    ----------
+    output : Optional[str, Path, TextIO]
```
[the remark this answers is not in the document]
oerc0122: It means that while the typehints are in the signature they're also in the docs, you end up with something that looks like:

> my_function(argument: dict) -> None:
>
> Parameters:
> - **argument** (_dict_) - Description of argument.
> 
> [...]

I think if you don't add the comment it doesn't add the bit in the parameters block. I do need to go through and remake those annotations more consistent and English rather than the automatic version my emacs plugin builds.

castep_outputs/cli/castep_outputs_main.py
```diff
@@ -1,5 +1,5 @@
 """
-Run main castep parser
+Main castep parser.
```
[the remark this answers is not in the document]
oerc0122: Well, I removed the verb because this is the module description, not a function. I do need to get better at writing this kind of documentation.

castep_outputs/parsers/__init__.py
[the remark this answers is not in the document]
oerc0122: cf `#:` here https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoproperty

castep_outputs/parsers/bands_file_parser.py
```diff
@@ -1,23 +1,63 @@
...
+    band: ThreeVector
+    #: List of eigenvalues for up component of band.
+    band_up: ThreeVector
+    #: List of eigenvalues for down component of band.
+    band_dn: ThreeVector
+    #: Position in space.
+    qpt: ThreeVector
+    #: Current spin component.
+    spin_comp: int
+    #: K point weight
```
oerc0122, line 33:
```suggestion
    #: K point weight.
```

castep_outputs/utilities/castep_res.py
oerc0122: Yeah, an awkward side effect of the `:meta hide-value:` unfortunately necessary for it to not *flood* the documentation with a 6 mile long regex. 

I will probably add a blank-line between the variables or if possible do you think it might be better to do:
```
INTNUMBER_RE = ... #: :meta hide_value:
```

castep_outputs/utilities/dumpers.py
oerc0122: Could do it as a see-also or link to the constant? Might be a good idea.

castep_outputs/cli/castep_outputs_main.py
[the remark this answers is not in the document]
ajjackson: In fairness I think this is a pretty good example of why the full stop shouldn't be mandated.

_Is_ it mandated for the module level docstring?

castep_outputs/utilities/castep_res.py
oerc0122: It was not possible to do the latter.

castep_outputs/parsers/__init__.py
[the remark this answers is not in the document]
ajjackson: Ah, that makes sense in terms of technically following the guidance.

Still looks weird. Maybe I'll get used to it.

castep_outputs/utilities/castep_res.py
ajjackson: I think a blank line after each regex will do the trick

castep_outputs/utilities/datatypes.py
oerc0122: I'm... not sure what this means, I need to ask someone who does.

castep_outputs/cli/castep_outputs_main.py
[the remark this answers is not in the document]
ajjackson: Euphonic does not always repeat the types in the docstring and they still get picked up from annotation and used in the parameter section of autodocs. In fact, you can see that in the places the type _was_ given in the docstring it appears twice! (But this may be because it wasn't formatted correctly.) e.g.

https://euphonic.readthedocs.io/en/stable/structure-factor.html#euphonic.structure_factor.StructureFactor

https://github.com/pace-neutrons/Euphonic/blob/master/euphonic/structure_factor.py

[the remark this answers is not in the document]
oerc0122: Huh. When I do that I lose the paren'd types:

> Parameters:
> - in_file – Input file to parse.
> - parser – Castep parser to use. If None will be determined from extension.
> - out_format – Format to dump as.
> - loglevel – Logging level.
> - testing – Whether used for test suite (disable processing fragile properties).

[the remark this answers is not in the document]
ajjackson: Might be a configuration option? I see Euphonic uses a `'sphinx_autodoc_typehints'` extension.

https://github.com/tox-dev/sphinx-autodoc-typehints

castep_outputs/utilities/datatypes.py
ajjackson: I can't find `shell_swp` or `shell_swp_end` in the Castep 24.1 source code. Where does this come from?

ajjackson: I don't see it in there; sounds like a question for Chris Pickard? Maybe that should be added to a TO-DO list somewhere and we can drop this docstring for now.

oerc0122: Yeah, there's also https://castep-docs.github.io/castep-docs/documentation/Pseudopotentials/otfg_string/ but I can't find it there either. I seem to remember it's some kind of debugging feature but I have no idea.

castep_outputs/cli/castep_outputs_main.py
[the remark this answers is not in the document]
ajjackson: I have opened an Issue for this, as the only (but large) outstanding change request. This PR is still valuable without it.

castep_outputs/parsers/__init__.py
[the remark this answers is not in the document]
ajjackson: Let us see if it becomes more tolerable over time 😆

ajjackson (approved):
That's all from me! Issue #169 opened to use a sphinx extension and remove duplicate type annotations from docstrings.

castep_outputs/cli/castep_outputs_main.py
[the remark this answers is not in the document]
oerc0122: Great, thanks. I'll probably do that in a separate PR.

PR #148, opened 2024-07-28
Refactor *_fmt file parsers

oerc0122:
Restructure *_fmt file parsers to improve code locality and use data names from file.

Originally part of #146


test/data_files/chdiff_fmt.json
```diff
@@ -41,7 +41,7 @@
       1
     ]
   ],
-  "chdiff": [
+  "charge": [
```
ajjackson, line 44: This looks like an improvement, when referring to the input file. But I don't quite understand where the old behaviour came from?

PR #151, opened 2024-07-30
Rewrite get_block to return oop style block

### LINKED ISSUE #149: Consider object-oriented Block
I don't like to recommend creating new classes where a Python built-in will do fine...

but looking at this return section of `get_block` (and the corresponding `out_fmt` parameter)

https://github.com/oerc0122/castep_outputs/blob/0512ad1edeed551de96caff831f1f79d5e3020d2/castep_outputs/utilities/castep_res.py#L47-L55

there is a strong code smell that this function should be consistently returning a Block object which provides the features or conversion to these other classes.

Specifically:
- `__str__` can be implemented to Do The Right Thing and receiving code modified from `get_block(**kwargs, out_fmt=str)` to `str(get_block(**kwargs))`
- `__bool__` can be implemented so `if block` behaves as though for an empty string
- `__iter__` can be implemented to allow idiomatic `list(get_block(**kwargs))` (where necessary: in most cases iterating over `block` will be sufficient.)
- a `strip_outer_lines()` method would improve readability elsewhere
- the `out_fmt=io.StringIO` option is currently unused; it could be dropped in favour of `io.StringIO(str(get_block(**kwargs)))` or constructed by a method `get_block(**kwargs).as_stringio()`

### LINKED ISSUE #152: Rename `cnt` arg
There _has_ to be a better name for this

_Originally posted by @ajjackson in https://github.com/oerc0122/castep_outputs/pull/151#discussion_r1698129032_

oerc0122:
Initial implementation of #149 .

No doubt needs further iteration in future PRs.

Fixes #149
Fixes #152

### DIFF
--- a/castep_outputs/utilities/castep_res.py
+++ b/castep_outputs/utilities/castep_res.py
@@ -2,12 +2,9 @@
 
 import itertools
 import re
-from typing import Dict, List, Sequence, TextIO, Union
+from typing import Dict, List, Sequence, Union
 
 from .constants import FST_D, MINIMISERS, SHELLS
-from .filewrapper import Block, FileWrapper
-
-# pylint: disable=too-many-arguments
 
 
 Pattern = Union[str, re.Pattern]
@@ -16,71 +13,6 @@ Pattern = Union[str, re.Pattern]
 def get_numbers(line: str) -> List[str]:
     """ Get all numbers in a string as a list """
     return NUMBER_RE.findall(line)
-
-
-def get_block(
-        init_line: str,
-        in_file: Union[TextIO, FileWrapper, Block],
-        start: Pattern,
-        end: Pattern,
-        *,
-        cnt: int = 1,
-        eof_possible: bool = False,
-) -> Block:
-    """
-    Check if line is the start of a block and return the block if it is.
-    Parameters
-    ----------
-    init_line : str
-        Initial line which may start the block.
-    in_file : ~typing.TextIO | FileWrapper
-        File handle to read data from.
-    start : Pattern
-        RegEx matched against `init_line` to see if is start of block.
-    end : Pattern
-        RegEx to verify if block has ended.
-    cnt : int
-        Number of times `end` must match before block is returned.
-    out_fmt : type
-        Datatype to return parsed block as.
-    eof_possible : bool
-        Whether it is possible block is ended by EOF.
-
-    Returns
-    -------
-    Block
-        Recovered block of data matching prereqs.
-
-    Notes
-    -----
-    Advances `in_file` as it does so.
-
-    Raises
-    ------
-    IOError
-        If EOF reached and ``not eof_possible``.
-    """
-
-    block = Block(in_file)
-
-    if not re.search(start, init_line):
-        return block
-
-    block += init_line
-    fnd = cnt
-    for line in in_file:
-        block += line
-        if re.search(end, line):
-            fnd -= 1
-            if fnd == 0:
-                break
-    else:
-        if not eof_possible:
-            if hasattr(in_file, 'name'):
-                raise IOError(f"Unexpected end of file in {in_file.name}.")
-            raise IOError("Unexpected end of file.")
-
-    return block
 
 
 def labelled_floats(labels: Sequence[str], counts: Sequence[Union[int, str, None]] = (None,),
--- a/castep_outputs/utilities/castep_res.py
+++ b/castep_outputs/utilities/castep_res.py
@@ -1,12 +1,11 @@
 """ Module containing all regexes """
 
-import io
 import itertools
 import re
 from typing import Dict, List, Sequence, TextIO, Union
 
 from .constants import FST_D, MINIMISERS, SHELLS
-from .filewrapper import FileWrapper
+from .filewrapper import Block, FileWrapper
 
 # pylint: disable=too-many-arguments
 
@@ -19,18 +18,55 @@ def get_numbers(line: str) -> List[str]:
     return NUMBER_RE.findall(line)
 
 
-def get_block(init_line: str, in_file: Union[TextIO, FileWrapper],
-              start: Pattern, end: Pattern, *, cnt: int = 1,
-              out_fmt: type = io.StringIO, eof_possible: bool = False):
-    """ Check if line is the start of a block and return
-    the block if it is, moving in_file forward as it does so """
+def get_block(
+        init_line: str,
+        in_file: Union[TextIO, FileWrapper, Block],
+        start: Pattern,
+        end: Pattern,
+        *,
+        cnt: int = 1,
+        eof_possible: bool = False,
+) -> Block:
+    """
+    Check if line is the start of a block and return the block if it is.
+    Parameters
+    ----------
+    init_line : str
+        Initial line which may start the block.
+    in_file : ~typing.TextIO | FileWrapper
+        File handle to read data from.
+    start : Pattern
+        RegEx matched against `init_line` to see if is start of block.
+    end : Pattern
+        RegEx to verify if block has ended.
+    cnt : int
+        Number of times `end` must match before block is returned.
+    out_fmt : type
+        Datatype to return parsed block as.
+    eof_possible : bool
+        Whether it is possible block is ended by EOF.
 
-    block = ""
+    Returns
+    -------
+    Block
+        Recovered block of data matching prereqs.
+
+    Notes
+    -----
+    Advances `in_file` as it does so.
+
+    Raises
+    ------
+    IOError
+        If EOF reached and ``not eof_possible``.
+    """
+
+    block = Block(in_file)
 
     if not re.search(start, init_line):
         return block
 
-    block = init_line
+    block += init_line
     fnd = cnt
     for line in in_file:
         block += line
@@ -44,15 +80,7 @@ def get_block(init_line: str, in_file: Union[TextIO, FileWrapper],
                 raise IOError(f"Unexpected end of file in {in_file.name}.")
             raise IOError("Unexpected end of file.")
 
-    if not block:
-        return ""
-    if out_fmt is str:
-        return block
-    if out_fmt is list:
-        return block.splitlines()
-    if out_fmt is io.StringIO:
-        return io.StringIO(block)
-    return out_fmt(block)
+    return block
 
 
 def labelled_floats(labels: Sequence[str], counts: Sequence[Union[int, str, None]] = (None,),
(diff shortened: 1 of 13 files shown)


castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +31,178 @@ def __next__(self):
...
+
+class Block:
+    """
+    Data block class returned from :func:`get_block`.
+
+    Emulates the properties of both a file, and sequence.
+    """
+    def __init__(self, parent: Union[TextIO, FileWrapper, "Block"]):
+        if isinstance(parent, (FileWrapper, Block)):
+            self._lineno = parent.lineno
```
ajjackson, line 94: No fallback, so `self._lineno` is undefined for TextIO input and `lineno` property will blow up. Would 0 be a reasonable fallback value?

castep_outputs/utilities/castep_res.py
```diff
@@ -19,18 +18,55 @@ def get_numbers(line: str) -> List[str]:
...
+    ----------
+    init_line : str
+        Initial line which may start the block.
+    in_file : ~typing.TextIO | FileWrapper
+        File handle to read data from.
+    start : Pattern
+        RegEx matched against `init_line` to see if is start of block.
+    end : Pattern
+        RegEx to verify if block has ended.
+    cnt : int
```
ajjackson, line 42: There _has_ to be a better name for this

castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +31,178 @@ def __next__(self):
...
+
+        Returns
+        -------
+        str
+            Name if name is known otherwise "unknown".
+        """
+        return self._name
+
+    @property
+    def lineno(self) -> int:
```
ajjackson, line 187: I find this word hard to scan. Is `line_number` really too verbose?

castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +31,178 @@ def __next__(self):
...
+
+    Emulates the properties of both a file, and sequence.
+    """
+    def __init__(self, parent: Union[TextIO, FileWrapper, "Block"]):
+        if isinstance(parent, (FileWrapper, Block)):
+            self._lineno = parent.lineno
+
+        self._name = parent.name if hasattr(parent, 'name') else 'unknown'
+        self._i = -1
+        self._data = []
```
ajjackson, line 98: It feels a bit error-prone that this class is responsible for getting the line number from parent, but not for collecting the lines that correspond to it. It would be very easy to construct a Block from a file, then skip past a header, _then_ start appending things to the Block. This would cause the line numbers to be out of sync.

Indeed, at any point one can append an out-of-sequence line to the Block, making line numbers incorrect.

Perhaps the real problem is the append (`__plus__`) method, and data should be provided directly in `__init__`, making the calling code responsible for providing a contiguous set of lines.

castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +31,178 @@ def __next__(self):
...
+
+        Parameters
+        ----------
+        fore : bool
+            Whether to strip leading line from data.
+        back : bool
+            Whether to strip trailing line from data.
+        """
+        self._data = self._data[fore:
+                                -back if back else None]
```
ajjackson, line 115: This should also increment the line number?

ajjackson (commented):
Feature set looks good and it seems to be cleaning up the places I'd hoped.

I'm not quite sure about the API and line-number handling at this stage, it seems fragile.

castep_outputs/utilities/filewrapper.py
oerc0122: This is inherited based on Python stdlib.
https://docs.python.org/3/library/fileinput.html#fileinput.lineno

ajjackson: Ah, fair enough then

castep_outputs/utilities/filewrapper.py
oerc0122: The line number and name are intended purely for errorhandling purposes to report, e.g. "Error on line (30) of 'MyFile.castep'". They aren't meant to be mainly accessible. Thus `_lineno` is just an offset into that file.

castep_outputs/utilities/castep_res.py
ajjackson: `end_count` or `n_end`, maybe?

castep_outputs/utilities/filewrapper.py
ajjackson: It is _especially_ important that the line number is correct if it is intended for debugging

ajjackson: I see two potential "safe" ways of doing it:
- Block is quasi-immutable, and should always be constructed with an initial lineno and some contiguous line data. (Stripping from the ends can be done correctly, but removing material from the middle or adding to the end can never be done safely if we assume that line-number = initial line + block line.)
- Block stores each line with a corresponding number, and manages these when it is mutated. Typically lines are appended with an `append_line(source: TextIO|FileWrapper|Block)` method that will get the current line _and its number_ from the source. We no longer care if the data is contiguous, the correct line number can still be obtained.

oerc0122: I see what you mean now. Yes, that's a good idea.

oerc0122: The other way which I was thinking of was making the Block quasi-immutable and making `get_block` a classmethod constructor for a block.

ajjackson: I did wonder about that; I think at this point it's ok for `get_block` to be a function or a classmethod. It _should_ become a classmethod if Block will be subclassed further down the line, needing a `get_block` that uses some different implementations of methods. But that doesn't seem likely enough right now to dominate the decision: go with what looks prettiest!

castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +34,290 @@ def __next__(self):
...
+
+        block = cls(in_file)
+
+        if not re.search(start, init_line):
+            return block
+
+        data: List[str] = []
+        data.append(init_line)
+
+        fnd = n_end
```
ajjackson, line 202: Very confusing that `fnd` ("found"?) starts high and counts down. `n_to_find`, maybe. Probably clearer to have `n_found` and count up.

castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +34,290 @@ def __next__(self):
...
+        Raises
+        ------
+        IOError
+            If EOF reached and ``not eof_possible``.
+        """
+
+        block = cls(in_file)
+
+        if not re.search(start, init_line):
+            return block
```
ajjackson, line 197: Can we put a comment above this line that reads something like `# Pattern not found, return zero-length block`? At first glance it looks like we're returning some data, which would be weird.

castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +34,290 @@ def __next__(self):
...
+        else:
+            if not eof_possible:
+                if hasattr(in_file, 'name'):
+                    raise IOError(f"Unexpected end of file in {in_file.name}.")
+                raise IOError("Unexpected end of file.")
+
+        block._data = tuple(data)
+        return block
+
+    def remove_bounds(self, fore: int = 1, back: int = 2):
```
ajjackson, line 218: The opposite of "fore" is, of course, "aft".

Which would be a fun pair of parameters but needlessly confusing. Maybe "front"/"back" or "start"/"end".

Or maybe fun is ok 🏴‍☠️

castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +34,290 @@ def __next__(self):
...
+        Returns
+        -------
+        list[str]
+            Block as list of lines.
+        """
+        return self._data
+
+    def rewind(self):
+        """
+        Rewind file to previous line.
```
ajjackson, line 260: This only moves the position marker and doesn't change the contained data. Could that cause some issues? Is that a good thing? Should we say something about it here?

castep_outputs/utilities/filewrapper.py
```diff
@@ -26,23 +34,290 @@ def __next__(self):
...
+
+        Useful for blocks not terminated by a clear
+        statement, where we can only check if next line
+        does *NOT* match.
+        """
+        self._i -= 1
+
+    def __bool__(self):
+        return bool(self._data)
```
ajjackson, line 269: ❤️

ajjackson (commented):
Looking nice! And API getting a bit more intuitive as we go. Still a few points of possible confusion commented on here.

castep_outputs/utilities/filewrapper.py
oerc0122: Not 100% convinced by that argument. C.f. foreground/background (not aftground, although maybe I should adopt it)
Maybe make it positional only? That possibly makes it more confusing.
 `before`, `after` obviously valid.

castep_outputs/utilities/filewrapper.py
oerc0122: Good point, copy of one above.

Probably also want to limit it to not rewinding indefinitely although thanks to the magic of Python getting the -5th element should work up to a whole second cycle round.

castep_outputs/utilities/filewrapper.py
```diff
@@ -253,16 +254,20 @@ def aslist(self) -> List[str]:
         list[str]
             Block as list of lines.
         """
-        return self._data
+        return list(self._data)
```
ajjackson, line 257: Is this in case some deviant decides to modify self._data?

oerc0122: It's to conform to the expectations of a list rather than a tuple. Especially given the function is `aslist`.

castep_outputs/utilities/filewrapper.py
ajjackson: Ok, it can stay if you're still happy with it 😄

oerc0122: Ended up making it positional only. Not sure if this is necessarily the wisest thing, but a reasonable IDE will identify the info and it's used that way anyway.

ajjackson (approved):
I think this is mergeable.

I note that the test coverage is very high, but there are no explicit unit tests on the Block class. It might be helpful to have those for troubleshooting in future.

PR #153, opened 2024-07-31
Add delocalised coordinates parsers

### LINKED ISSUE #111: Parse delocalised geomopt
`Geometry/Constraints/benchmark.out.castep-23.1.castep-python-1.0.python-3.6.9.inp=Si8-nonlinear.param`

oerc0122:
Fixes #111

### DIFF
--- a/castep_outputs/parsers/castep_file_parser.py
+++ b/castep_outputs/parsers/castep_file_parser.py
@@ -26,11 +26,14 @@ from ..utilities.datatypes import (
     CellInfo,
     CharTable,
     ConstraintsReport,
+    DelocActiveSpace,
+    DelocInternalsTable,
     DipoleTable,
     ElasticProperties,
     FinalConfig,
     GeomTable,
     InitialSpin,
+    InternalConstraints,
     KPointsList,
     KPointsSpec,
     MDInfo,
@@ -1088,6 +1091,38 @@ def parse_castep_file(castep_file_in: TextIO,
             logger("Found %s geom_block", typ)
 
             curr_run["minimisation"].append(_process_geom_table(block))
+
+        # GeomOpt Deloc
+
+        elif block := Block.from_re(line, castep_file,
+                                    "INTERNAL CONSTRAINTS", REs.EMPTY, n_end=2):
+
+            if Filters.GEOM_OPT not in to_parse:
+                continue
+
+            curr_run["internal_constraints"] = _process_internal_constraints(block)
+
+        elif block := Block.from_re(line, castep_file,
+                                    "Message: Generating deloc", "Message: Generation of deloc"):
+
+            if Filters.GEOM_OPT not in to_parse:
+                continue
+
+            logger("Found Delocalisaed GeomOpt Table")
+
+            curr_run.setdefault("delocalised_internal", {})
+            curr_run["delocalised_internal"].update(_process_deloc_table(block))
+
+        elif block := Block.from_re(line, castep_file,
+                                    "The size of active space", REs.EMPTY):
+
+            if Filters.GEOM_OPT not in to_parse:
+                continue
+
+            logger("Found Delocalisaed GeomOpt")
+
+            curr_run.setdefault("delocalised_internal", {})
+            curr_run["delocalised_internal"].update(_process_deloc_act_space_table(block))
 
         # TDDFT
         elif block := Block.from_re(line, castep_file,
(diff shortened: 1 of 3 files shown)


castep_outputs/parsers/castep_file_parser.py
```diff
@@ -2403,3 +2438,68 @@ def _process_elastic_properties(block: Block) -> ElasticProperties:
...
+    for line in block:
+        if not line.strip():
+            continue
+
+        _, type_, target, actual, *definitions = line.split()
+        if actual == "Satisfied":
+            target, actual = actual, target
+        else:
+            target = float(target)
+        actual = float(actual)
```
ajjackson, line 2459: Some fun stuff is happening here but without input examples it's hard to understand how these variable names relate to something that can be two floats, or a string and float in the other order.

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -2403,3 +2438,68 @@ def _process_elastic_properties(block: Block) -> ElasticProperties:
...
+
+        definitions = " ".join(definitions).split(")")
+        const = {val[0]: tuple(val[1:])
+                 for definition in definitions
+                 if definition and
+                 (val := to_type(get_numbers(definition), int))}
+        assert type_ in ("Bond", "Angle", "Torsion")
```
ajjackson, line 2466: Should this assert be above `if actual == "Satisfied":` (i.e. directly below `type_` is assigned)?

castep_outputs/parsers/castep_file_parser.py
oerc0122: So this is due to weird and awkward formatting coming out of CASTEP, I don't understand it either.

Rough translation to Python:
```python
#19) vali (in) : current value of the constrained internal
#20) rc (in) : prescribed value of the constrained internal

# Unsatisfied
print(f"{index:d3} Bond  {rc:f9.3}{vali:f10.3} ... # Constraints listing")

#Satisfied
print(f"{index:d3} Bond  {vali:f9.3} Satisfied ... # Constraints listing")
```
So it swaps when satisfied for some reason. I will try to work out a better expression of it.

ajjackson: Maybe this is a case where trying two different regex would be clearer than splitting to (potentially incorrect) variable names?

oerc0122: At the moment, I thankfully avoid any regex, I think probably a comment would be more clear. Having a think.

ajjackson: Perhaps the real problem here is stuffing "Satisfied" into "target" when it is a difference piece of information from the target value.

oerc0122: I did consider renaming it "actual_or_sat", but I felt that added less clarity. I have renamed it to "current" rather than "actual" and added a (hopefully) clarifying comment.

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -2403,3 +2438,71 @@ def _process_elastic_properties(block: Block) -> ElasticProperties:
...
+        _, type_, target, current, *definitions = line.split()
+        assert type_ in ("Bond", "Angle", "Torsion")
+
+        if current == "Satisfied":
+            # Castep dumps current in target spot for reasons..?
+            # index type current "Satisfied" ...
+            target, current = current, float(target)
+        else:
+            # index type target current ...
+            target, current = float(target), float(current)
```
ajjackson, line 2463: This works for me. The remaining ugliness reflects the underlying behaviour 😅

castep_outputs/parsers/castep_file_parser.py
ajjackson: An alternative approach would be to add a separate "satisfied" boolean to the data object, and set target to `None` for these blocks.

This will make it clearer to a downstream parser what is going on, and we don't have to worry any more about people doing number things to a string. (They might do them to None, but at least a noisy error is more likely.)

oerc0122: That is a very good idea, going to do that. 

How's this for a brain-bender then?
```python
if type == "bond":
    print(f"{j:d3} Bond   {rc:f9.*3*}{vali:10.*3*} ...")
elif type == "torsion":
   print(f"{j:d3} Torsion{rc:f9.*2*}{vali:10.*2*} ...")
```
"angle" constraints are similar to "torsion" ones.
I'm glad I'm not doing exact REs (though I "should" since there's strictly no spacing for very large values, but if you've got a million angstrom long constraint you have other problems)

ajjackson: I guess bonds just have to be a _bit_ more precise!

How did so many academic Fortran developers learn to never leave spaces between number fields?

castep_outputs/utilities/datatypes.py
```diff
@@ -116,6 +116,44 @@ class GeomTable(TypedDict):
...
 
+class InternalConstraints(TypedDict):
+    """Internal constraints information."""
+
+    #: Constraint type.
+    type: Literal["Bond", "Angle", "Torsion"]
+    #: Current value of constraint.
+    actual: float
+    #: Required constraint value.
+    target: Literal["Satisfied"] | float
```
ajjackson, line 127: I think this now needs splitting to separate fields to match the code above?

castep_outputs/utilities/datatypes.py
```diff
@@ -116,6 +116,44 @@ class GeomTable(TypedDict):
...
+    """Table of delocalised internal coordinates."""
+
+    #: Number of bond coordinates.
+    num_bonds: int
+    #: Number of angular coordinates.
+    num_angles: int
+    #: Number of dihedral coordinates.
+    num_dihedrals: int
+    #: Total number of internal coordinates.
+    num_inversions: int
```
ajjackson, line 142: Should this be "num_internals?"

ajjackson (changes requested):
A couple of issue with the schema.

Which presumably means it isn't being tested anywhere? But it is hard to validate a TypedDict except by running mypy. Maybe this beartype feature would help, I've never tried it: https://beartype.readthedocs.io/en/latest/faq/#how-do-i-only-type-check-while-running-my-test-suite

(The requested change is to update the schema, not to add more CI.)

oerc0122:
> A couple of issue with the schema.
> 
> Which presumably means it isn't being tested anywhere? But it is hard to validate a TypedDict except by running mypy. Maybe this beartype feature would help, I've never tried it: https://beartype.readthedocs.io/en/latest/faq/#how-do-i-only-type-check-while-running-my-test-suite
> 
> (The requested change is to update the schema, not to add more CI.)

You are correct, and highlighted the fact that I'd broken my `flycheck` configuration and it wasn't running `mypy` (my hacked together `flycheck-ruff` accidentally prevented any other flycheck running on Python). There are quite a few `mypy` errors, so I think I'll fix those as a separate PR, if you're happy I've fixed these for now. 

Eventually, I suspect I'll add `mypy` (or similar) to the CI.

ajjackson (approved):

PR #163, opened 2024-08-15
Fix annotations through Ruff 

oerc0122:


ajjackson:
I'm surprised that ruff likes the "implicit optional" `def func(x: SomeType = None)` syntax? It isn't even mentioned in the docs for `typing`: https://docs.python.org/3/library/typing.html And in PEP-484

> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

 https://peps.python.org/pep-0484/#union-types

ajjackson (approved):
All looks nice and clean. Thanks for separating this so cleanly from other work, this made manual review viable.

ajjackson (changes requested):
Hmm, conflict here between `ruff` desire to use `type|type` and the package support for Python <3.10.

Why does it need to support 3.8/3.9 again?

oerc0122:
Strangely despite setting ruff's target-version to `py38`, it still seems convinced that this is for the better and valid.

oerc0122:
> Hmm, conflict here between `ruff` desire to use `type|type` and the package support for Python <3.10.
> 
> Why does it need to support 3.8/3.9 again?

I think dropping 3.8/3.9 comes later. At which point I will run the cleanup again.

ajjackson:
I still see lots of pipe characters in the diff, how are the tests passing?

oerc0122:
`from __future__ import annotations` fixes the issue for the annotations only. So they're happy with `|`. The cases which were failing were in-code (e.g. `typing.cast(int | float, var)` or `MyType = dict[str,  str]`)

ajjackson (dismissed):
Ah, ok then!

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -1777,19 +1780,19 @@ def _process_qdata(qdata: Dict[str, Union[str, List[str]]]) -> QData:
     return cast(QData, qdata)
 
 
-def _parse_magres_block(task: int, inp: Block) -> Dict[Union[str, AtomIndex],
-                                                       Union[str, Dict[str, Optional[float]]]]:
+def _parse_magres_block(task: int, inp: Block) -> dict[str | AtomIndex,
+                                                       str | dict[str, float | None]]:
```
ajjackson, line 1784: Oof, there are a lot of possible input type combinations here. How many are tested?

ajjackson, line 1784: Oof, there is a pretty big configurational space of output types here; presumably corresponding to an equivalent variety of input data. How well is it tested?

ajjackson (approved):
I don't see any surprises following the change in target, it still seems fine.

oerc0122:
> I'm still not quite convinced that universally changing `Optional` to `|None` is a good move, but this seems to be a bit of a controversial topic [astral-sh/ruff#4858](https://github.com/astral-sh/ruff/issues/4858). I suspect a lot of people find Optional confusing and are glad to ditch it. I find it a bit more expressive as long as it is _only_ used in parameters 🤷
> 
> Ditching Union is good and `ruff` currently doesn't help us do that without ditching Optional as well. So let's give it a try.

I do agree. `|` is better than union and `Optional[X]` is more informative than `X` if more verbose. If/when Ruff updates to allow `|` with `Optional` maybe we'll switch back.

Honestly (I know, heresy), I prefer implicit `Optional` IFF the default is `None` and the value is set in the function to a "valid" value.

ajjackson:
Yeah, I didn't mind implicit Optional either 😆  - with the same caveat that Optional should only be used when the `None` is some kind of fallback. If it's a "meaningful" value, `| None` _was_ a clearer expression of that intent.

PR #170, opened 2024-09-11
Add autodoc typehints; enable ruff pydocstyle

### LINKED ISSUE #169: Consider using sphinx_autodoc_typehints to avoid duplication between type hint and docstring
This came out of review for #146 . If developers are encouraged to use the autodoc documentation, I don't see any reason not to use it. The only downside is that it is sometimes convenient to use `help()` in interactive sessions: but this should be weighted against the risk that the docstring is incorrect.

https://github.com/tox-dev/sphinx-autodoc-typehints

oerc0122:
Fixes #169

### DIFF
--- a/castep_outputs/parsers/castep_file_parser.py
+++ b/castep_outputs/parsers/castep_file_parser.py
@@ -1,7 +1,9 @@
 # pylint: disable=too-many-lines, too-many-branches, too-many-statements, too-many-locals
 """
-Extract results from .castep file for comparison and further processing
+Extract results from .castep file for comparison and further processing.
 
+Notes
+-----
 Port of extract_results.pl
 """
 from __future__ import annotations
@@ -79,9 +81,8 @@ from .xrd_sf_file_parser import parse_xrd_sf_file
 
 
 class Filters(Flag):
-    """
-    Enum of possible filters for CASTEP file parsing
-    """
+    """Enum of possible filters for CASTEP file parsing."""
+
     BS = auto()
     CELL = auto()
     CHEM_SHIELDING = auto()
@@ -137,7 +138,26 @@ class Filters(Flag):
 
 def parse_castep_file(castep_file_in: TextIO,
                       filters: Filters = Filters.HIGH) -> list[dict[str, Any]]:
-    """ Parse castep file into lists of dicts ready to JSONise """
+    """
+    Parse castep file into lists of dicts ready to JSONise.
+
+    Parameters
+    ----------
+    castep_file_in
+        File to parse.
+    filters
+        Parameters to parse.
+
+    Returns
+    -------
+    list[dict[str, Any]]
+        Parsed data.
+
+    Raises
+    ------
+    ValueError
+        On invalid top-level blocks.
+    """
     # pylint: disable=redefined-outer-name
 
     runs: list[dict[str, Any]] = []
@@ -1438,7 +1458,7 @@ def _process_elf(block: Block) -> list[float]:
 
 
 def _process_hirshfeld(block: Block) -> dict[AtomIndex, float]:
-    """ Process Hirshfeld block to dict of charges """
+    """Process Hirshfeld block to dict of charges."""
     return {
         atreg_to_index(match): float(match["charge"]) for line in block
         if (match := re.match(rf"\s+{REs.ATREG}\s+(?P<charge>{REs.FNUMBER_RE})", line))
@@ -1446,7 +1466,7 @@ def _process_hirshfeld(block: Block) -> dict[AtomIndex, float]:
 
 
 def _process_thermodynamics(block: Block) -> Thermodynamics:
-    """ Process a thermodynamics block into a dict of lists """
+    """Process a thermodynamics block into a dict of lists."""
     accum: Thermodynamics = defaultdict(list)
     for line in block:
         if "Zero-point energy" in line:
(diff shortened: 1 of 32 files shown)


castep_outputs/parsers/castep_file_parser.py
```diff
@@ -137,7 +138,26 @@ class Filters(Flag):
...
+    Parameters
+    ----------
+    castep_file_in
+        File to parse.
+    filters
+        Parameters to parse.
+
+    Returns
+    -------
+    list[dict[str, Any]]
```
ajjackson, line 153: Should this be here?

(There are a few explicit return types given in this diff.)

ajjackson (commented):
Looking good! And looks like you were already _quite_ well aligned with Pydocstyle :-)

Settings look fine, no objection to letting magic methods go docstring-free as these already have a well-defined purpose.

I guess I'd better try to build these docs and take a look, next.

ajjackson:
Observations from building the docs:
- The type hints seem to be showing up ok
- Where there is a redundant return type, it isn't getting duplicated in docs
- In practice the API docs mostly consist of "private" methods. Should we be showing those at same prominence as public methods? (In fact they are listed first...)

castep_outputs/parsers/castep_file_parser.py
oerc0122: I think the return type is still a helpful indicator for describing the output. I would prefer to leave those.

oerc0122: Only if you do the weird thing of:
```rest

    Returns
    -------
        Parsed data.
```
Apparently

ajjackson: I don't think it's necessary. Look at the `to_dict()` method of QpointPhononModes: return (or parameter) section in the docstring, and we still get a return type in the generated docs.

https://euphonic.readthedocs.io/en/stable/qpoint-phonon-modes.html#euphonic.qpoint_phonon_modes.QpointPhononModes.to_dict

https://github.com/pace-neutrons/Euphonic/blob/826888122ee93e1191de7afc324f541c6d85c464/euphonic/qpoint_phonon_modes.py#L606

oerc0122: Yes, but in order to get the description (which admittedly is not very descriptive), it has to be randomly indented 4 spaces, which looks weird. While I'm not sure that description is useful *here*, it is elsewhere and would look either very weird or inconsistent.

ajjackson: It looks less weird if you name it. The docstring for calculate_pdos (on the same class) has

```
        Returns
        -------

        dos
            A collection of spectra, with the energy bins on the x-axis
            and PDOS for each atom in the unit cell on the y-axis,
            metadata describes the index and species of each PDOS. If
            the PDOS is not neutron-weighted, it is in units of modes
            per energy unit per atom, such that the integrated area of
            the total DOS is equal to 3. If the PDOS is weighted by the
            neutron cross-section, it is in units of area per unit
            energy per atom, see Notes for details.
```

Which looks ok in the auto docs (return type is given separately but that seems fine) https://euphonic.readthedocs.io/en/stable/qpoint-phonon-modes.html#euphonic.qpoint_phonon_modes.QpointPhononModes.calculate_pdos

oerc0122: Oh, it lets you get away with that? Yeah, that might work where descriptions are worthwhile!

ajjackson: I don't think it's so much "get away with that" as "how it is supposed to be used"?

ajjackson:
I found the relevant config in *docs/_config.yml*: 

```
    autodoc_default_options:
      members: True
      inherited-members: False
      private-members: True
      show-inheritance: True
```
I suggest setting private-members to False and inherited-members to True, so the full public (and only public) API is shown in docs.

oerc0122:
> I found the relevant config in _docs/_config.yml_:
> 
> ```
>     autodoc_default_options:
>       members: True
>       inherited-members: False
>       private-members: True
>       show-inheritance: True
> ```
> 
> I suggest setting private-members to False and inherited-members to True, so the full public (and only public) API is shown in docs.

Hm... showing `inherited-members` means that all the `TypedDict` s have all dictionary methods documented which is really ugly. Given we don't *really* have inheritance in castep_outputs (yet, not saying it'll never happen), I think it's better left off until we decide that inheriting methods is important.

ajjackson:
Ugh, never mind `inherited-members`, then! Maybe there is a way to get more granular control but I didn't find it in a few minutes of looking...

castep_outputs/parsers/castep_file_parser.py
oerc0122: So because of the template we're currently using this does lead to awkwardly confusing structure, e.g.:
```
castep_outputs.utilities.utility.normalise_string(string)
Normalise a string.

This includes:
- Removing leading/trailing whitespace.
- Making all spacing single-space.

Parameters:
string (str) – String to process.

Returns:
Normalised string.

Return type:
out
```
I'm not against changing templates, but until we do, I think these are better left as-is.

ajjackson: I don't quite follow. This is the rendered output for docs, right? It looks fine apart from the Return type being "out", which looks totally incorrect. Is it not correctly parsing the return type from the function signature?

ajjackson (approved):
Happy for this to be merged in its current state; we can quibble about return value conventions at a later date if it is causing trouble.

castep_outputs/parsers/castep_file_parser.py
oerc0122: No the case is you either have to have:
```
    Returns
    -------
        Parsed data.
```
Which looks weird,
```
    Returns
    -------
    out
        Parsed data.
```
(Based on what you said above re: `dos`) Which sets "return type" to `out`

or fully document the return type as I am doing.

ajjackson: I see. It works for Euphonic, so basically this template is broken. Maybe open an Issue to look at that some other day at very low priority?

PR #172, opened 2024-09-30
Fix for comment lines inside blocks

pointedsphere:
Hi Jacob,

I'm Scott, one of Matt's PhD students that was looking at incorporating castep_outputs into some example scripts.

Found a small bug with the reading of a cell file where there are comments within a block that contain numbers. A bit of a niche thing, but the CASTEP output of final post geom-opt cells (with write_cell_structure) includes comments of this form.

This PR is to account for this possibility. Unfortunately cannot use `re.split('[#!]',line)[0]` to get the bit of the line without a comment as the end of the lines `Kinetic eigenvalue #   1 =     0.19929544E+02` in CASTEP output files are also thrown away by this. As such, this just discards full comment lines.

Thanks,

Scott


oerc0122:
This seems like a far-reaching and potentially harmful change to automatically ignore any/all `#`/`!` characters in any file parsed by castep_outputs. 

I would suggest instead a separate `strip_comments` `filter` in the utilities. 

```python
comment_re = re.compile("^\s*#")
itertools.filterfalse(comment_re.match, my_data)
```

pointedsphere:
I agree it is a bit drastic.

I'm not 100% on exactly how you were suggesting to add the filter. I've modified the PR to use `itertools.filterfalse` in the block reading to ignore full line comments. However, if I've missed the mark very happy to close the commit and leave it as an issue in the tracker.

castep_outputs/utilities/filewrapper.py
```diff
@@ -203,7 +204,7 @@ def from_re(
         data.append(init_line)
 
         found = 0
-        for line in in_file:
+        for line in filterfalse(COMMENT_LINE_RE.match,in_file):
```
oerc0122, line 207: Like I said, I don't think this should be applied generally in FileWrapper, but as a utility function when parsing chunks of data where comments should be stripped.

oerc0122:
> I agree it is a bit drastic.
> 
> I'm not 100% on exactly how you were suggesting to add the filter. I've modified the PR to use `itertools.filterfalse` in the block reading to ignore full line comments. However, if I've missed the mark very happy to close the commit and leave it as an issue in the tracker.

@pointedsphere c.f. #173  and check that meets your needs.

PR #173, opened 2024-10-01
Add comment stripping and apply when parsing cell/param files

oerc0122:
Add comment stripping and apply when parsing .cell/.geom blocks.


castep_outputs/utilities/utility.py
```diff
@@ -548,3 +550,150 @@ def fix_data_types(in_dict: MutableMapping, type_dict: dict[str, type]):
...
+
+    def strip(data):
+        for line in data:
+            line = comment_re.split(line)[0].rstrip()
+            if not line:
+                continue
+
+            yield line
+
+    yield from strip(data)
```
ajjackson, line 602: I don't follow why a `yield from` is used here. The coroutine-priming property is not used. Why not inline the whole `strip` function and yield directly from that loop?

castep_outputs/utilities/utility.py
```diff
@@ -548,3 +550,150 @@ def fix_data_types(in_dict: MutableMapping, type_dict: dict[str, type]):
...
+    ... End of line # comment
+    ... ''')
+    >>> '|'.join(_strip_initial_comments(inp, comment_char={"#",}))
+    'Hello|End of line # comment'
+    """
+    comment_re = re.compile(rf"^\s*[{''.join(comment_char)}]")
+    data = filterfalse(comment_re.match, data)
+    data = map(str.rstrip, data)
+    data = filter(None, data)
+    yield from data
```
ajjackson, line 645: This `yield from` is ok as shorthand, I think: we can easily see it is just unpacking from `filter` and don't need to go looking for coroutine complications.

If we had `toolz.pipe` it could be used here, but the multiple `data` assignments are easy to read and understand so it really isn't needed.

castep_outputs/utilities/utility.py
```diff
@@ -548,3 +550,150 @@ def fix_data_types(in_dict: MutableMapping, type_dict: dict[str, type]):
...
+    >>> from io import StringIO
+    >>> inp = StringIO('''
+    ... Hello
+    ... # Initial line comment
+    ... End of line # comment
+    ... ''')
+    >>> '|'.join(_strip_inline_comments(inp, comment_char={"#",}))
+    'Hello|End of line'
+    """
+    comment_re = re.compile(f"[{''.join(comment_char)}]")
```
ajjackson, line 592: A non-obvious consequence of this is that the comment_char _must_ be an individual character. Multi-character tokens such as `<--` will not work correctly when placed in the set. Perhaps that should be documented somewhere.

castep_outputs/utilities/utility.py
```diff
@@ -548,3 +550,150 @@ def fix_data_types(in_dict: MutableMapping, type_dict: dict[str, type]):
...
+    comment_re = re.compile(rf"^\s*[{''.join(comment_char)}]")
+    data = filterfalse(comment_re.match, data)
+    data = map(str.rstrip, data)
+    data = filter(None, data)
+    yield from data
+
+def strip_comments(
+        data: TextIO | FileWrapper | Block,
+        *,
+        comment_char: str | set[str] = "#!",
```
ajjackson, line 650: comment_char is not documented in the Parameters section

castep_outputs/utilities/utility.py
```diff
@@ -548,3 +550,150 @@ def fix_data_types(in_dict: MutableMapping, type_dict: dict[str, type]):
...
+    >>> x = strip_comments(inp, remove_inline=True)
+    >>> '|'.join(x)
+    'Hello|End of line'
+    """
+    if not isinstance(comment_char, set):
+        comment_char = set(comment_char)
+
+    stripped_comments = (_strip_inline_comments(data, comment_char=comment_char)
+                         if remove_inline else
+                         _strip_initial_comments(data, comment_char=comment_char))
```
ajjackson, line 697: Alternatively, with less interleaving of information:
```python
strip_function = _strip_inline_comments if remove_inline else _strip_initial_comments
stripped_comments = strip_function(data, comment_char=comment_char)
```

ajjackson (commented):
Looks pretty nice! A few queries.

castep_outputs/utilities/utility.py
oerc0122: Vestigial from the original formulation. Good catch.

castep_outputs/utilities/utility.py
oerc0122: Hmm... good point, I hadn't considered that. It's an easy enough fix though.

castep_outputs/utilities/utility.py
oerc0122: Done and with the note to sort the multiple-character comments.

ajjackson (approved):
🚀

PR #175, opened 2024-11-07
Update readme to be more accurate and link to docs

oerc0122:


README.rst
```diff
@@ -82,30 +82,31 @@ using the ``PyYAML`` engine if available and the ``RUAMEL`` engine if not.
...
-If you need a specific parser rather than by extension it is possible to pass
-them as the second argument, or call them directly.
+If you need a specific parser rather than determining it by extension
+it is possible to pass them as the second argument, or call them
+directly.
 
 ::
 
-   import castep_outputs
+   import castep_outputs as co
```
ajjackson, line 104: `co` isn't very greppable, maybe we can establish a better idiom here.

(It's a fairly common digraph and collides with "count", "correction", "correlation", "constraint", "complex" ... https://irp.fas.org/doddir/army/fm34-40-2/appa.pdf)

ajjackson (approved):
A modest and clear improvement. Just one non-blocking comment to consider.

README.rst
oerc0122: Interesting article. Not seen that one before.

It was more that I wanted to not recommend importing `*`, and also highlight those functions which are castep_outputs without filling the README with `castep_outputs.`. 

In my own code I'd do (as I do below this)
```python
from castep_outputs import parse_single
```
but this doesn't serve to highlight those features which are from castep_outputs.

Any suggestions?

Other possible ideas might be:
- `ctp_out` (still too long?)
- `c_o` (which I dislike)

PR #176, opened 2024-11-28
Change md/geom parsing to frame function for easier use.

oerc0122:
- Fix issue with reading all frames of MD/Geom files
- Add a function to parse a single frame of an MD calculation from a block enabling generator approaches.
- Makes `Block.__bool__` verify it contains non-whitespace data.


castep_outputs/parsers/md_geom_file_parser.py
```diff
@@ -67,47 +68,65 @@ class MDGeomTimestepInfo(TypedDict, total=False):
...
-            curr = defaultdict(list)
-            curr["ions"] = {}
+
+    for line in block:
+        if not line.strip():
+            pass
         elif not TAG_RE.search(line):  # Timestep
             curr["time"] = to_type(get_numbers(line)[0], float)
 
         elif match := ATDATTAG.match(line):
```
ajjackson, line 94: Is this "AT_DAT_TAG"? Took a while to parse...

castep_outputs/parsers/md_geom_file_parser.py
```diff
@@ -67,47 +68,65 @@ class MDGeomTimestepInfo(TypedDict, total=False):
...
             ion = atreg_to_index(match)
-            if ion not in curr["ions"]:
-                curr["ions"][ion] = {}
+            curr["ions"].setdefault(ion, {})
             curr["ions"][ion][match.group("tag")] = to_type([match.group(d) for d in FST_D], float)
 
         elif match := TAG_RE.search(line):
             curr[match.group("tag")].append([*to_type(get_numbers(line), float)])
 
+    if curr and curr["ions"]:
```
ajjackson, line 102: It looks to me like `curr` always contains "ions" here and so is always Truth-y?
Indeed, `curr["ions"]` is guaranteed to be a dict so it should always be safe to iterate over `.values()`; maybe this `if` statement can be removed.

castep_outputs/parsers/md_geom_file_parser.py
oerc0122: It is `ATomicDATaTAG`, yeah. from a time before naming.

castep_outputs/parsers/md_geom_file_parser.py
oerc0122: I think the `return curr` used to be in the block and has been shuffled in some refactor or other. I'll take a look and have a think.

castep_outputs/parsers/md_geom_file_parser.py
```diff
@@ -67,47 +68,65 @@ class MDGeomTimestepInfo(TypedDict, total=False):
...
+    list[MDGeomTimestepInfo]
+        Step-by-step Parsed info.
+    """
+    while "END header" not in md_geom_file.readline():
+        pass
+    md_geom_file.readline()
+    steps = []
+    while str(block := Block.from_re("", md_geom_file, "", "^$", eof_possible=True)).strip():
+        steps.append(parse_md_geom_frame(block))
+
```
ajjackson, line 129: Would it be useful to pack `bool(str(block).strip())` into a `.__bool__()` method on Block so that "False-y" Blocks can be checked directly?

castep_outputs/parsers/md_geom_file_parser.py
ajjackson: Is that an established abbreviation from CASTEP source, or a new one for this repo?

ajjackson (commented):
A couple of tidy-up/clarity suggestions but nothing too critical.

castep_outputs/parsers/md_geom_file_parser.py
oerc0122: No, a nightmare dredged up while in seas of RegEx. I will fix it.

castep_outputs/parsers/md_geom_file_parser.py
oerc0122: `bool(block) == bool(block._data)` so if there's an empty string in there it still counts. Maybe a `has_text` method.

ajjackson: Is the current behaviour useful or could we define all whitespace-only blocks as false-y?

oerc0122: That would be a possibility.

castep_outputs/parsers/md_geom_file_parser.py
oerc0122: Moved to #180, #181

ajjackson:
I think `if curr and curr["ions"]` is the only thing left, now

ajjackson (changes requested):
```
castep_outputs/utilities/dumpers.py:53:5: D403 [*] First word of the docstring should be capitalized: `yaml` -> `Yaml`
   |
52 |   def yaml_dumper(data: Any, file: TextIO):
53 |       """
   |  _____^
54 | |     yaml format dumper.
55 | | 
56 | |     Parameters
57 | |     ----------
58 | |     data
59 | |         Data to dump.
60 | |     file
61 | |         File to dump to.
62 | |     """
   | |_______^ D403
63 |       yaml.dump(data, file)
   |
   = help: Capitalize `yaml` to `Yaml`

Found 1 error.
[*] 1 fixable with the `--fix` option.
```

Not ruff's finest moment 😞 . I guess we actually want `YAML format dumper.`

The full stop is also grammatically dubious (yet required).

ajjackson (approved):

PR #177, opened 2024-11-28
Fix some miscellaneous issues breaking expectations/tests

oerc0122:
- Accidentally labelled `parse_args` empty example as doctest causing failure with `pytest --doctest-modules`
- Force `output` to `Path` for required method in `parse_single`
- Make `Block.get_lines` add lines as elements of tuple rather than as characters.
- Fix issue with latest `ruff` version.
- Pin `ruff` version to avoid this problem in future

### DIFF
--- a/castep_outputs/cli/args.py
+++ b/castep_outputs/cli/args.py
@@ -61,7 +61,7 @@ def parse_args(to_parse: Sequence[str] = ()) -> argparse.Namespace:
 
     Examples
     --------
-    >>> parse_args()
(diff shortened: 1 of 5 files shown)


ajjackson (approved):
These seem fine and I don't mind them being batched as they are small and non-interactive.

Pinning an _exact_ ruff version seems a bit aggressive, but I suppose as developers are not obliged to use the `[lint]` extra locally it's not the end of the world.

PR #183, opened 2025-01-20
Fix EOF issue.

oerc0122:
Some (task farm) runs do not output peak memory, just run to EOF/first empty.


ajjackson (approved):
Looks fine:
- logic checks out
- style is consistent
- relevant test data added
- expected output is in the test reference json
- nothing else was broken

PR #186, opened 2025-01-22
Improve typing in cell-param parser

oerc0122:
- Improve type descriptions in cell/param parser
- Change positions info labels to catch meaning.
- Simplify logic of devel block parsing.
- Add better units handling to `positions_(abs|frac)` block
- Add test on `positions_abs` with units


castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -77,48 +152,58 @@ def _parse_devel_code_block(in_block: Block) -> dict[str, str | float | int]:
...
     -------
-    dict[str, str | float | int]
+    dict[str, str | float]
         Parsed info.
     """
-    in_block = str(in_block)
-    in_block = " ".join(in_block.splitlines())
-    matches = re.finditer(REs.DEVEL_CODE_BLOCK_GENERIC_RE, in_block, re.IGNORECASE | re.MULTILINE)
-    devel_code_parsed = {}
+    main_block = " ".join(map(str.strip, in_block))
```
ajjackson, line 158: Am I understanding this correctly: the lines in a devel block are always joined into one? Does CASTEP not have meaningful newlines in devel blocks?

ajjackson:
Apart from above query, this all looks reasonable 🚀

castep_outputs/parsers/cell_param_file_parser.py
oerc0122: No, I don't believe so. I'll ask.

oerc0122: > Jacob Wilkins  14:49
> Quick question on Devel Codes, are there any devel codes with meaningful newlines? In castep-outputs, I just parse devel code as though it's all one line, but that may be incorrect.

> krefson 14:51
> No, I think it's all treated as whitespace.

ajjackson (approved):

PR #190, opened 2025-02-07
Refactor testing

### LINKED ISSUE #178: test_dumper passes with ruamel.yaml but not pyyaml

### LINKED ISSUE #184: Move testing over to pytest
While `castep_outputs` itself should be stdlib, there is no real reason why the testing framework needs to be. Migrating to `pytest` should hopefully produce better errors and nicer syntax, particularly in `test_dumpers` as we need to add PyYAML testing (#178).

oerc0122:
- Migrate to pytest testing (Fixes #184)
- Move tests out of main folder
- Add coverage reporting to github
- Add testing for pyyaml (Fixes #178)
- Refactor how ruamel vs pyyaml is implemented


test/test_dumpers.py
```diff
@@ -0,0 +1,86 @@
...
+    elif out_format == "ruamel":
+        module = pytest.importorskip("ruamel.yaml")
+        yaml_eng = module.YAML(typ="unsafe")
+        loader = yaml_eng.load
+    elif out_format == "pyyaml":
+        module = pytest.importorskip("yaml")
+        try:
+            from yaml import CLoader as Loader
+        except ImportError:
+            from yaml import Loader
```
ajjackson, line 57: Version 6 of pyyaml has wheels on PyPI; these presumably provide CLoader. Do we really need this fallback for the test suite?

test/test_dumpers.py
```diff
@@ -0,0 +1,86 @@
...
+    # Check produced data files identical
+    test = io.StringIO()
+    parse_all(test, out_format=out_format, **{typ: [file]})
+    test.seek(0)
+
+    comp_dict = loader(test)
+    comp_dict = normalise(comp_dict, {dict: _to_complex})
+    assert comp_dict == ref_dict
+
+    # Check contents identical
```
ajjackson, line 78:
```suggestion
    # Check parse_single (i.e. wrapper) gives same result as specific loader
```

test/test_filters.py
```diff
@@ -0,0 +1,42 @@
...
+))
+def test_filter_all(filters, keys):
+    with _TEST_FILE.open("r", encoding="utf-8") as file:
+        data = parse_castep_file(file, filters=filters)
+    assert keys <= data[0].keys()
+
+def test_filter_none():
+    with _TEST_FILE.open("r", encoding="utf-8") as file:
+        data = parse_castep_file(file, filters=Filters.NONE)
+    assert not data
```
ajjackson, line 42: Other tests have an `if __name__ = '__main__'` block but not this one.

Are they being used?

ajjackson (commented):
Well this took a lot less time to look at than the diff suggested 😆

test/test_dumpers.py
oerc0122: I doubt it, but it's very little visual noise for the safety. Guess I could do:
```suggestion
        module = pytest.importorskip("yaml")
        Loader = module.CLoader if hasattr(module, "CLoader") else module.Loader
```

test/test_filters.py
oerc0122:
```suggestion
    assert not data


if __name__ == "__main__":
    pytest.main()
```

test/test_dumpers.py
ajjackson: If it's important to have the tests work with an old or source version of pyyaml, then leave as-is.

Otherwise you might as well
`Loader = module.CLoader` and let the `NameError` happen.

oerc0122: Does seem like `Loader` and `CLoader` have (slightly) different behaviour. And that some builds don't include CLoader (as of 2020?)

https://stackoverflow.com/questions/47715566/cannot-load-cloader-with-pyyaml/62543781#62543781

ajjackson: Happy to leave as-is then

PR #196, opened 2025-02-10
Add polarisation parser

### LINKED ISSUE #157: Add parser for polarisation
`Test/Electronic/Polarisation/benchmark.out.castep-23.1.castep-python-1.0.python-3.6.9.inp=pbtio3-polar-ncp.param`

oerc0122:
Add parser for berry phase calculation. 

Unfortunately a lot of special casing due to multiple dumps of the same info and weird tables and judicious ignorance of repeated info from verbose output.

Fixes #157

### DIFF
--- a/castep_outputs/parsers/castep_file_parser.py
+++ b/castep_outputs/parsers/castep_file_parser.py
@@ -103,6 +103,7 @@ class Filters(Flag):
     OPTICS = auto()
     PARAMETERS = auto()
     PHONON = auto()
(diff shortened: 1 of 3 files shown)


ajjackson (approved):
The code looks about as complicated as necessary for the problem at hand, and the test looks correct.

I wondered about passing through the knowledge that you have a "verbose" block rather than re-inferring it, but as there are other block types to infer in the same loop the current way seems cleaner.

PR #209, opened 2025-02-12
Ignore if coverage report fails

oerc0122:
Forked repos don't necessarily have the tokens for bot. 

Ignore failures on coverage reporting as not essential.


ajjackson:
That's quite an interesting limitation, actually. I think this action will _only_ work from a pull request event and can't be manually dispatched. So how _would_ you see the coverage of an external contribution? Is there no other way around this?

oerc0122:
The coverage report will be in the test report, just can't be posted to the PR without tokens it seems.

ajjackson:
I see! That seems tolerable.
Screenshot 2025-02-12 at 09 58 31

ajjackson (approved):

PR #211, opened 2025-03-18
Update workflows and metadata

oerc0122:
- Add testing and metadata for Python 3.11,3.12 (for 3.13 c.f #212)
- Use Ruff action
  - Fix resulting issues
- Use latest action versions

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -11,4 +11,4 @@ jobs:
       with:
         src: "./castep_outputs"
         args: "check"
-        version: "0.11.x"
+        version: "0.11.0"
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -11,3 +11,4 @@ jobs:
       with:
         src: "./castep_outputs"
         args: "check"
+        version: "0.11.x"
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -5,19 +5,9 @@ on: [push]
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.10"]
     steps:
-    - uses: actions/checkout@v3
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v3
+    - uses: actions/checkout@v4
+    - uses: astral-sh/ruff-action@v3
       with:
-        python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install ruff
-    - name: Analysing the code with Ruff
-      run: |
-        ruff check castep_outputs
+        src: "./castep_outputs"
+        args: "check"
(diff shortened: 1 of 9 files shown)


.github/workflows/ruff.yml
```diff
@@ -5,19 +5,10 @@ on: [push]
...
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install ruff
-    - name: Analysing the code with Ruff
-      run: |
-        ruff check castep_outputs
+        src: "./castep_outputs"
+        args: "check"
+        version: "0.11.x"
```
oerc0122, line 14:
```suggestion
        version: "0.11.0"
```

.github/workflows/ruff.yml
```diff
@@ -5,19 +5,10 @@ on: [push]
...
-        python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install ruff
-    - name: Analysing the code with Ruff
-      run: |
-        ruff check castep_outputs
+        src: "./castep_outputs"
+        args: "check"
```
ajjackson, line 13: "Check" is the default, but no harm supplying it.

Is there a particular reason not to run ruff at the project folder level? (i.e. the default.)

.github/workflows/test_castep_outputs.yml
```diff
@@ -22,12 +22,12 @@ jobs:
       pull-requests: write
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
ajjackson, line 25: 3.8 is fully EOL now, do you need to support it?

pyproject.toml
```diff
@@ -25,6 +25,8 @@ classifiers = [
   "Programming Language :: Python :: 3.8",
```
ajjackson, line 25:
```suggestion
```

.github/workflows/ruff.yml
oerc0122: Mostly just because it catches `test`/`doc` otherwise.

ajjackson (approved):
This seems clean and non-controversial except for the continued Python 3.8 support.

.github/workflows/ruff.yml
ajjackson: As long as you have _considered_ catching test and consciously decided against it, ok

.github/workflows/test_castep_outputs.yml
oerc0122: The problem is HPC machines aren't always the most up-to-date.

ajjackson: There are plenty of ways around that these days, all my cluster accounts have a `micromamba` on them...

If not following Python's own EOL schedule, is there some other schedule to check against?

ajjackson: (Anyway it's not a blocker for this PR, I already approved 😄 )

PR #213, opened 2025-03-21
Use `fractions.Fraction` for fractional representation parsing

oerc0122:
Removes maintenance burden and as of 3.12 supports extended things.

c.f.
https://docs.python.org/3/library/fractions.html#fractions.Fraction


ajjackson (approved):
This is a nice improvement on what is there.

Seems a slight shame to immediately convert to float and lose some round-trip perfection, but I get that this is needed for e.g. JSON dumping. Maybe in a future version...

PR #214, opened 2025-03-21
Add colourisation to pytest output

oerc0122:
Force coloured output to colourise pytest output.


ajjackson (approved):
Looks good in the web interface, no obvious downside as only CI is affected

PR #216, opened 2025-04-09
Add vermin to CI to check python versions

oerc0122:
Ensure that minimum version is supported by quick CI check before testing.


ajjackson (approved):
Implementation seems simple enough and it's running ok. Not too much to review here!

In principle one could be clever about linking the minimum version to some other central location, but really the worst that will happen is CI fails and then we know to update it.

PR #217, opened 2025-04-11
Add script for predicting generated files from cell/param info.

### LINKED ISSUE #215: Get files from cell/param
Use a parsed cell/param file to get a list of expected outputs.

oerc0122:
- Add script to predict generated files (Fixes #215).
- Moves `castep_file_parser.process_pspot_string` to `cell_param_file_parser.parse_pspot_string`
- Adds parsing for `xc_definition` blocks to `cell_param_file_parser`

Following discussions, the scripting interface will be placed here and the CLI interface moved to `castep_outputs_tools`


castep_outputs/scripts/get_generated_files.py
```diff
@@ -0,0 +1,488 @@
...
+    KeyError
+        Impossible combination.
+    """
+    out_files = set()
+    devel_code = param_data.get("devel_code", {})
+    raw_task = param_data.get("task", "SINGLEPOINT").upper()
+    spectral_theory = SpectralTheory(param_data.get("spectral_theory", "dft"))
+    spectral_task = SpectralTask(param_data.get("spectral_task", "bandstructure"))
+
+    out_files.add(f"{seedname}-out.cell")
```
ajjackson, line 126: Huh, today I learned that spectral calculations automatically enable `write_cell_structure`. *Why*?

And why does it ignore my `write_cell_structure: False`?

castep_outputs/scripts/get_generated_files.py
```diff
@@ -0,0 +1,488 @@
...
+    if param_file is None:
+        param_file = seedname.with_suffix(".param")
+    if cell_file is None:
+        cell_file = seedname.with_suffix(".cell")
+
+    param = parse_cell_param_file(param_file)[0] if param_file.exists() else {}
+    cell = parse_cell_param_file(cell_file)[0] if cell_file.exists() else {}
+    return get_generated_files(param, cell, seedname.with_suffix(""))
+
+@get_generated_files.register(dict)
```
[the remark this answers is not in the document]
oerc0122: Not sure how well `functools.singledispatch` works with specific subtypes of generics. I know it doesn't work well with concrete generics. I can try, I suppose.

castep_outputs/scripts/get_generated_files.py
```diff
@@ -0,0 +1,488 @@
...
+def _(
+        param_data: CellParamData, cell_data: CellParamData, src: Path | str = "seedname",
+) -> list[str]:
+    seedname = Path(src).stem
+    param_data = deepcopy(param_data)
+    devel_code = param_data.get("devel_code", {})
+    raw_task = param_data.get("task", "SINGLEPOINT").upper()
+    task = Task(raw_task)
+    xc = param_data.get("xc_functional", "LDA").upper()
```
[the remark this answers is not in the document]
oerc0122: I always forget that exists. Will need to work out how `xc_definition` plays with setting internal CASTEP properties.

castep_outputs/scripts/get_generated_files.py
oerc0122: Question for Keith?

castep_outputs/scripts/get_generated_files.py
```diff
@@ -0,0 +1,488 @@
...
+    out_files.add(f"{seedname}-out.cell")
+
+    if param_data.get("spectral_restart", True):
+        out_files.add(f"{seedname}.*.spec")
+
+    if param_data.get("write_orbitals") or raw_task == "BANDSTRUCTURE":
+        out_files.add(f"{seedname}.orbitals")
+    if spectral_theory is SpectralTheory.TDDFT:
+        if spectral_task is SpectralTask.CORELOSS:
+            raise KeyError("Invalid param file")
```
oerc0122, line 135:
```suggestion
```

castep_outputs/scripts/get_generated_files.py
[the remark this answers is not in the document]
ajjackson: No idea, fine to use `dict` if it doesn't work.

castep_outputs/scripts/get_generated_files.py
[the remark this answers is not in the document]
oerc0122: Should be better?

castep_outputs/scripts/get_generated_files.py
oerc0122: Raised an issue with castep.

castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -387,6 +444,64 @@ def _parse_symops(block: Block) -> list[dict[str, ThreeByThreeMatrix | ThreeVect
...
+            if key == "nlxc_screening_function":
+                block_data["params"][key] = val
+            if key == "nlxc_ppd_int":
+                block_data["params"][key] = val.upper() == "ON"
+            if key == "nlxc_divergence_corr":
+                block_data["params"][key] = val.upper() == "ON"
+
+            continue
+
+        block_data["xc"][key] = float(val)
```
ajjackson, line 482: I'm not sure this nested structure with a `continue` is more readable than

```suggestion
        if key == "nlxc_screening_length":
            block_data["params"][key] = float(val)
        elif key == "nlxc_screening_function":
            block_data["params"][key] = val
        elif key in ("nlxc_ppd_int", "nlxc_divergence_corr"):
            block_data["params"][key] = val.upper() == "ON"
        else:
            block_data["xc"][key] = float(val)

```

castep_outputs/tools/get_generated_files.py
```diff
@@ -0,0 +1,504 @@
...
+    devel_code = param_data.get("devel_code", {})
+    raw_task = param_data.get("task", "SINGLEPOINT").upper()
+    task = Task(raw_task)
+
+    xc_f = param_data.get("xc_functional", "LDA").upper()
+    xc_d = param_data.get("xc_definition")
+
+    xc = xc_d["xc"].keys() if xc_d else {xc_f}
+
+    is_mgga = xc & {"ms2", "rscan", "r2scan", "libxc_mgga", "libxc_hyb_mgga"}
```
ajjackson, line 235: Won't this fail because in the libxc case `xc` is something like `{"libxc_mgga_xc_tpsslyp1w"}`? Where is the string split?

castep_outputs/parsers/cell_param_file_parser.py
oerc0122: Indeed, part of overly frequent restructuring. Thanks for catching this.

castep_outputs/scripts/get_generated_files.py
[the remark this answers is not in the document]
oerc0122:
```
E   TypeError: Invalid first argument to `register()`: dict[str, typing.Union[str, float, tuple[float, str], dict[str, typing.Any], dict[typing.Union[str, tuple[str, int]], typing.Union[str, dict[str, float]]], dict[str, typing.Union[collections.abc.Sequence[typing.Union[str, float, dict[str, typing.Union[str, float]]]], str, float, dict[str, typing.Union[str, float]], dict[str, typing.Union[collections.abc.Sequence[typing.Union[str, float, dict[str, typing.Union[str, float]]]], str, float, dict[str, typing.Union[str, float]]]]]]]]. Use either `@register(some_class)` or plain `@register` on an annotated function.
```

ajjackson:
That looks like it should work! I think we just need to add a libxc / xc_definition case to the unit test.

ajjackson:
ruff error looks legit if rather pedantic

castep_outputs/tools/get_generated_files.py
oerc0122: The string is split in `get_xc_info`

test/test_cell_param_parser.py
```diff
@@ -504,6 +506,58 @@ def test_devel_code(self):
                                                     "int_value": 31,
                                                     "string_value": "Hello"}})
 
+    def test_xc_def(self):
+        test_text = io.StringIO("""
+%block xc_definition
+oep 1.0
+hf 2.1
+lda 0.1
+libxc_mgga_xc 2.
```
ajjackson, line 515:
```suggestion
libxc_mgga_xc_zlp 2.
```

castep_outputs/tools/get_generated_files.py
ajjackson: Ah, got it! Might be clearer if shown in the test of that function.

test/test_cell_param_parser.py
ajjackson: I see. In that case it would still be good to include the full XC name (in both input and output) of this test, so that it is more realistic and doesn't _look_ like it would have been trimmed off already.

test/test_cell_param_parser.py
```diff
@@ -504,6 +506,58 @@ def test_devel_code(self):
...
+%endblock xc_definition
+""")
+        test_dict = parse_cell_param_file(test_text)[0]
+        pprint.pprint(test_dict)
+        self.assertEqual(test_dict, {"xc_definition": {
+            "xc": {
+                "oep": 1.0,
+                "hf": 2.1,
+                "lda": 0.1,
+                "libxc_mgga_xc": 2.0,
```
oerc0122, line 529:
```suggestion
                "libxc_mgga_xc_zlp": 2.0,
```

test/test_cell_param_parser.py
oerc0122: If castep doesn't like any lowercasing do you think this is a bug in castep? Inputs are supposed to be (mostly) case insensitive where reasonable.

ajjackson:
Am I right in thinking there is no pathway in the test suite that checks a libxc definition will be interpreted correctly? All the *test_get_generated_files* cases use `xc_functional`.

test/test_get_generated_files.py
```diff
@@ -0,0 +1,251 @@
...
+            },
+        },
+    },
+     {
+         "task": "singlepoint",
+         "metals_method": "dm",
+         "opt_strategy": "memory",
+         "xc_definition": {
+             "xc": {
+                 "LIBXC_GGA_XC_ZLP": 1.0,
```
ajjackson, line 92:
```suggestion
                 "LIBXC_GGA_XC_B97_D": 1.0,
```
ZLP is a meta-GGA, I guess you wanted to go GGA to get the PBE pseudopotential outputs!

ajjackson (approved):

PR #218, opened 2025-05-01
Move `MDGeomParser` into castep_outputs

oerc0122:


ajjackson (commented):
This seems useful and nicely written!

I think I see a design issue, however, which can be solved with better separation of the "data container" and "generator" aspects of the object.

ajjackson (approved):
I think all my comments have been addressed, except for the new one about broken IndexError logic 😄

ajjackson (changes requested):
Just need to fix the IndexError

castep_outputs/tools/md_geom_parser.py
```diff
@@ -0,0 +1,220 @@
...
+        -------
+        MDGeomTimestepInfo
+            Parsed frame.
+
+        Raises
+        ------
+        IndexError
+            Requested frame out of range.
+        """
+        if frame not in range(-len(self), len(self)):
```
ajjackson, line 121: This implementation is pretty, but the cost scales linearly with data size. And this method is used by `__iter__` so the cost of iterating has gone quadratic rather than linear.

I appreciate it's a small overhead in practice, but it feels bad 😖

oerc0122: Which part do you think the scales linearly? `in range` is a constant-time operation, as is creating the range. It's slightly more expensive since it involves a subtract, mod against the step and a less than in practice, but I think the clarity outweighs that given it's unlikely to be a high performance operation.

ajjackson: Aha! Did not know `range()` supported constant-time `in`, that's quite handy. This is fine, then.

ajjackson (approved):
🚀

PR #219, opened 2025-05-02
Improve jupyter book docs

oerc0122:
- Add actual documentation on `castep_outputs`' usage.
- Make parser aliases public
- Allow passing string file extension to `parse_single`


pyproject.toml
```diff
@@ -51,6 +51,8 @@ version = {attr = "castep_outputs.__version__"}
 line-length = 100
 indent-width = 4
 target-version = "py39"
+include = ["*.py", "*.pyi", "**/pyproject.toml"]
```
ajjackson, line 54: Why lint .pyi files when the project doesn't ship any?

ajjackson:
Aside from the error situation this seems like a nice example. It's not a full-fledged beginner tutorial, but it's a useful point to dive in for intermediates. And that is really the target audience for now!

oerc0122:
> I think this block is not working as intended: I get ValueError rather then the customised TypeError

At some point I'd changed the error types to be more pythonic and this had sat with the old one.

ajjackson (approved):
Seems to work now

PR #221, opened 2025-05-02
Deprecate Python 3.8 and associated fixes

oerc0122:
Python 3.8 is EoL and limits functionality.

(also stop complaints)

### DIFF
--- a/.github/workflows/test_castep_outputs.yml
+++ b/.github/workflows/test_castep_outputs.yml
@@ -37,7 +37,7 @@ jobs:
       pull-requests: write
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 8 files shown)


ajjackson:
Any idea why Vermin thinks this is still valid Python 3.8 code?

oerc0122:
> Any idea why Vermin thinks this is still valid Python 3.8 code?

My best guess is naive handling of `from __future__ import annotations` as making all types viable (even in assignment). 

At the end of the day, I see it as a safeguard against obvious issues (and debugger in some cases) more than a strict rule. It's true that false negatives happen, but the tests should reveal those.

ajjackson:
Ah yes, `--eval-annotations` bumps it up to 3.9

ajjackson (approved):
Changes look correct to me. Ruff is happy, vermin is happy (with `--eval-annotations` it hits 3.9 as minimum).

PR #222, opened 2025-05-02
Add more linting

oerc0122:
Enable more ruff checks and fix the associated files.

### DIFF
--- a/castep_outputs/cli/castep_outputs_main.py
+++ b/castep_outputs/cli/castep_outputs_main.py
@@ -90,7 +90,7 @@ def parse_all(
         *,
         loglevel: int = logging.WARNING,
         testing: bool = False,
-        **files,
+        **files: Sequence[Path],
 ) -> None:
     """
     Parse all files in files dict.
@@ -121,7 +121,6 @@ def parse_all(
 
     if output is None:
         file_dumper(data, sys.stdout)
-        print()
     elif isinstance(output, io.TextIOBase):
         file_dumper(data, output)
     else:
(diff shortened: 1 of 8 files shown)


castep_outputs/processing/processing.py
```diff
@@ -1,9 +1,10 @@
 """Module to assist processing of parsed outputs."""
 
-from typing import NamedTuple
+from typing import NamedTuple, TypeVar
 
 from ..utilities.castep_res import get_atom_parts
 
+Self = TypeVar("Self", bound="AtomLabel")
```
ajjackson, line 7: This name could be a problem if you need more than one in the same file... but that can be dealt with as needed.

castep_outputs/cli/castep_outputs_main.py
```diff
@@ -121,7 +121,6 @@ def parse_all(
 
     if output is None:
         file_dumper(data, sys.stdout)
-        print()
```
ajjackson, line 124: This removes an empty line from the output. (But that was already an inconsistency with the other dumper mode.)

Is this expected/ok?

ajjackson (approved):
All looks fine, except for one (non-blocking) question.

castep_outputs/processing/processing.py
oerc0122: That's exactly the plan. But I can't use `typing_extensions` as external dependency. I wish Python made `typing_extensions` stdlib like `__future__`, but it's an experimental place too.

castep_outputs/cli/castep_outputs_main.py
oerc0122: Yeah, there shouldn't have been any `print`s. Presumably this was leftover from when I was initially writing it.

PR #223, opened 2025-05-06
Enable extended ruff checks

oerc0122:
Some checks including `DOC` were not enabled without `preview = true`. 

Add `preview = true` and fix associated errors.
Fix issue with `Self` in `FileWrapper` having two definitions.

### DIFF
--- a/castep_outputs/cli/castep_outputs_main.py
+++ b/castep_outputs/cli/castep_outputs_main.py
@@ -48,7 +48,7 @@ def parse_single(in_file: str | Path | TextIO,
 
     Raises
     ------
-    KeyError
+    ValueError
         If invalid `parser` provided.
     """
     logging.basicConfig(format="%(levelname)s: %(message)s", level=loglevel)
@@ -60,11 +60,12 @@ def parse_single(in_file: str | Path | TextIO,
         ext = in_file.suffix.strip(".")
 
         if ext not in ALL_PARSERS:
-            raise KeyError(f"Parser for file {in_file} (assumed type: {ext}) not found")
+            raise ValueError(f"Parser for file {in_file} (assumed type: {ext}) not found")
 
         parser = ALL_PARSERS[ext]
 
-    assert parser is not None
+    if parser is None:
+        raise ValueError("Unable to determine parser. Please specify through arguments.")
 
     data = parser(in_file)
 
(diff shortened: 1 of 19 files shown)


castep_outputs/parsers/castep_file_parser.py
```diff
@@ -2288,7 +2464,7 @@ def _process_phonon(block: Block, logger: Logger) -> list[QData]:
             next(char_table)
             char: list[CharTable] = []
             for char_line in char_table:
-                if re.search("[-=]{4,}", char_line):
+                if re.search(r"[-=]{4,}", char_line):
```
ajjackson, line 2467: Presumably from https://docs.astral.sh/ruff/rules/unraw-re-pattern/

I'm not sure of the value in requiring raw strings for regex that don't use backslash? I guess it means you don't have to worry about checking/forgetting if you edit it later?

castep_outputs/parsers/magres_file_parser.py
```diff
@@ -62,9 +63,7 @@ class MagresInfo(TypedDict):
     calc_pspot: str
     calc_xcfunctional: str
     #: Crystal lattice.
-    lattice: tuple[float, float, float,
-                   float, float, float,
-                   float, float, float]
+    lattice: tuple[float, float, float, float, float, float, float, float, float]
```
ajjackson, line 66: This seems a downgrade but maybe worth it for consistency.

If this data was meant to be more structured, the lists would be nested, right?

castep_outputs/parsers/tddft_file_parser.py
```diff
@@ -91,7 +91,8 @@ def parse_tddft_file(tddft_file: TextIO) -> TDDFTFileInfo:
                     match["energy"] = float(match["energy"])
 
                     dip = list(to_type(get_numbers(match["dipoles"]), float))
-                    dip = [complex(real, imag) for real, imag in zip(dip[0::2], dip[1::2])]
+                    dip = list(map(complex, dip[0::2], dip[1::2]))
```
ajjackson, line 94: Nice

ajjackson (approved):
This all seems to be implemented fine and changes are tolerable.

I have highlighted a few where we might learn something from further discussion.

castep_outputs/cli/castep_outputs_main.py
```diff
@@ -72,7 +70,7 @@ def parse_single(in_file: str | Path | TextIO,
 
     if out_format == "json" or testing:
         data = normalise(data, {dict: json_safe, complex: json_safe})
-    elif out_format in ("yaml", "ruamel"):
+    elif out_format in {"yaml", "ruamel"}:
```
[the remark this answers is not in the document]
oerc0122: Yeaaaaah, at least it's consistent within itself.

ajjackson (approved):

PR #224, opened 2025-05-20
Fix version printing, currently always prints 0.1

oerc0122:


ajjackson (approved):
Tested locally, all good

PR #235, opened 2025-09-25
Deprecate Python 3.9 and fix ruff upgrade issues

oerc0122:


.github/workflows/test_castep_outputs.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       pull-requests: write
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12"]
```
ajjackson: Any thoughts about extending to test on higher versions?

PR #236, opened 2025-09-26
Make the MD/Geom reader slightly more defensive

oerc0122:
Adds more informative error (rather than trying to read file indefinitely and failing with EoF) and tries to handle cases where the MD/geom header is missing more elegantly.


test/test_md_geom_parser.py
```diff
@@ -1,112 +1,119 @@
...
+        parse_md_geom_file(test_text)
+
+
+def test_md_geom_no_header(caplog):
+    test_text = io.StringIO(FAKE_FILE.replace("END header", "").replace("BEGIN header", ""))
+
+    parse_md_geom_file(test_text)
+
+    for rec in caplog.records:
+        assert rec.message == "Non-standard md/geom file: missing header."
```
ajjackson, line 119: Very nice!

ajjackson (approved):

PR #241, opened 2026-01-28
Catch present devel-codes

oerc0122:
Catch devel codes which are specified by existing in devel blocks rather than having a value.

### DIFF
--- a/castep_outputs/parsers/cell_param_file_parser.py
+++ b/castep_outputs/parsers/cell_param_file_parser.py
@@ -5,9 +5,8 @@ from __future__ import annotations
 import logging
 import re
 from collections import Counter, defaultdict
-from collections.abc import Callable, Sequence
 from functools import partial
-from typing import Any, Literal, TextIO, TypedDict, cast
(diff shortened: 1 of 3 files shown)


castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -380,6 +383,15 @@ def _parse_devel_code_block(in_block: Block) -> DevelBlock:
 
         devel_code_parsed[key] = to_type(val, typ)
 
+    # Catch present components
+    for par in re.finditer(r"(?<![:=])\b[A-Za-z_-]+\b(?![:=])", main_block):
```
ajjackson: (Not tested)

```suggestion
    present_code_re = re.compile(r"""
        (?<![:=])        # Does not follow : or =
        \b[A-Za-z_-]+\b  # One "word"
        (?![:=])         # Not followed by : or =
    """, re.VERBOSE)

    for par in re.finditer(present_code_re, main_block):
```

ajjackson (approved):

PR #245, opened 2026-04-01
Port docs to sphinx and update actions

oerc0122:
- Port docs to sphinx
- Update actions to latest version
- Switch some optional dependencies to dev groups 
- Add docs build test

Fixes #243


ajjackson (approved):
It builds locally for me and looks fine. Perhaps it would be a good idea to workflow-dispatch and push to the public docs before final merge of this branch? Then there is a handy place already to fix problems...

PR #246, opened 2026-04-08
Add epme_bin parser

### LINKED ISSUE #204: Create `.epme_bin` parser

oerc0122:
Add parser for `.epme_bin` files.

Fixes #204


castep_outputs/bin_parsers/epme_bin_parser.py
```diff
@@ -0,0 +1,243 @@
...
+        Data destination.
+    """
+    _check_header(reader, b"ATOM")
+
+    n_ion = accum["n_ions"] = reader.get(int)
+    accum["real_lattice"] = reader.get(float)
+    accum["ions"] = {}
+
+    for ind, spec, pos in reader.get_dtype_cycle((int, str, float), n=n_ion):
+        accum["ions"][spec.strip(), ind] = pos
```
ajjackson, line 142: Is `pos` a float or some collection of `float`?

A similar thing happens further down with eigenvalues: from this part of the API it looks like we're fetching a scalar but presumably it is actually an array. It does make the calls nice and clean, but seems like it could cause confusion at some point 🤔

castep_outputs/bin_parsers/epme_bin_parser.py
```diff
@@ -0,0 +1,243 @@
...
+        "n_bands": int,
+        "min_band": int,
+        "eigenvalues": float,
+        "velocities": float,
+    }
+
+    for data in reader.get_dtype_cycle(dtypes, n=total):
+        ns = data.pop("ns")
+        nk = data.pop("nk")
+        accum["kpoints"][ns, nk] = data
```
ajjackson, line 174: Cute!

castep_outputs/bin_parsers/epme_bin_parser.py
```diff
@@ -0,0 +1,243 @@
...
+    _check_header(reader, b"EPCOUPLING")
+
+    n_kpoint_pairs = reader.get(int)
+    n_bands = isqrt(reader.get(int))
+
+    if "n_kpoint_pairs" in accum and n_kpoint_pairs != accum["n_kpoint_pairs"]:
+        raise ValueError("Number of kpoint pairs doesn't match between EPCOUPLING and KPOINTS.")
+
+    if "n_bands" in accum and n_bands != accum["n_bands"]:
+        raise ValueError("Number of bands doesn't match between EPCOUPLING and KPOINTS.")
```
ajjackson: Are the "not in accum yet" real cases that we can check are covered properly?

castep_outputs/bin_parsers/fortran_bin_parser.py
```diff
@@ -93,7 +95,7 @@ def skip(self, n: int, /) -> None:
                 size = int.from_bytes(self.file.read(4), "big")
                 self.file.seek(size + 4, SEEK_CUR)
 
-    def get(self, typ: type[T]) -> T:
+    def get(self, typ: type[T]) -> T | tuple[T, ...]:
```
ajjackson: This flexibility does make it harder to reason about what is coming back. Would it make sense to instead have e.g. `get_collection` which _always_ returns `tuple[T, ...]`? I don't expect there are many cases we should be flexible about the scalar/tuple nature of the return value.

ajjackson (commented):
The parser looks very clean for a parser, the infrastructure is paying off!

But I'm not sure about this adaptation of `get` to return tuples sometimes; it's a move away from the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle).

castep_outputs/bin_parsers/fortran_bin_parser.py
oerc0122: This is what the type signature has always been and this is simply correcting it. 

The issue is that each fortran unformatted record is:
`nbytes, data, nbytes`

The `data` can be many data or one datum (real, complex, int, ...), we don't know until we read it without first checking the size. 

Realistically, we only know based on the key that we're reading whether it "should" be multiple or one, however, I can't think of a nice notation to note that there should be many data. I could have something like:
```python
reader.get((int, ...))
```
rather than
```python
reader.get(int)
```
But to some extent here it's a matter of perspective on "we're processing the next record as ..." if that makes sense. 

`to_type` is what already handles the `T | tuple[T, ...]` side and it does that for both `str` and `bytes`, we could rework that.

castep_outputs/bin_parsers/epme_bin_parser.py
oerc0122: No, I'm just thinking if we call these functions externally from some other parser.

ajjackson: It's a private function at the moment

oerc0122: True, though it just seemed like a sensible check. I can remove it.

ajjackson: We can keep/test the `!= accum[...]` part of the logic. I'd just let the AttributeError happen if someone starts running them out of order.

castep_outputs/bin_parsers/fortran_bin_parser.py
ajjackson: Ok, I'd be happy to push back this discussion to another issue/time.

But it sounds like it would make sense to initially treat the binary data as _always_ returning a tuple, and explicitly cast some fields to scalar. i.e.

```python
    def get_dtype_cycle(
        self,
        dtypes: tuple[type[T], ...],
        *,
        n: int | None = None,
    ) -> Iterator[tuple[tuple[T, ...], ...]]:
...

    for data in reader.get_dtype_cycle(dtypes, n=total):
        ns = as_scalar(data.pop("ns"))
        nk = as_scalar(data.pop("nk"))
        data["n_bands"] = as_scalar(data["n_bands"])
        data["min_band"] = as_scalar(data["min_band"])

        accum["kpoints"][ns, nk] = data
```

where `as_scalar()` is something like [more_itertools.only](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.only)

(I appreciate that sketch isn't very DRY and can be improved further...)

ajjackson:
Ok, parser typing stuff is now its own Issue. I think the only outstanding concern is the untested `ValueError` cases.

ajjackson (dismissed):
Lovely!

castep_outputs/bin_parsers/epme_bin_parser.py
```diff
@@ -0,0 +1,243 @@
...
+        Data destination.
+    """
+    _check_header(reader, b"ATOM")
+
+    n_ion = accum["n_ions"] = reader.get(int)
+    accum["real_lattice"] = reader.get((float, ...))
+    accum["ions"] = {}
+
+    for ind, spec, pos in reader.get_dtype_cycle((int, str, (float, ...)), n=n_ion):
+        accum["ions"][spec.strip(), ind] = pos
```
ajjackson, line 142: This is much clearer with the new type assignments 😄

castep_outputs/bin_parsers/fortran_bin_parser.py
```diff
@@ -112,34 +113,56 @@ def get(self, typ):
...
             Loaded data.
+
+        See Also
+        --------
+        get_dtype_iter :
+        get_dtype_dict :
         """
-        while True:
-            yield tuple(self.get(dtype) for dtype in dtypes)
+        ind = count() if n is None else range(n)
```
ajjackson, line 162: This is a pretty cool way of doing "sometimes infinite"

ajjackson (dismissed):
Ok, I think this is good to go!

(I guess you should sort out the linting though. Maybe blacklist COM812, that one fights with `ruff format`.)

ajjackson (dismissed):

ajjackson:
I _assume_ these are just rebases, so the approval stands

PR #247, opened 2026-04-08
Fix some miscellaneous typing issues

oerc0122:


castep_outputs/utilities/filewrapper.py
```diff
@@ -21,12 +22,14 @@ class FileWrapper:
         File to wrap and control.
     """
 
-    def __init__(self, file: TextIO) -> FWSelf:
+    Self = TypeVar("Self", bound="FileWrapper")
```
ajjackson, line 25: Huh, today I learned something about class variable scoping. Function signatures can see them without using `self` or `cls`! Interesting.

```
In [13]: class Thing:
    ...:     hello = 4
    ...:     def oh(self, x: int = hello) -> int:
    ...:         return x
    ...:     def my(self) -> int:
    ...:         return hello
    ...:

In [14]: Thing().oh()
Out[14]: 4

In [15]: Thing().my()
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[15], line 1
----> 1 Thing().my()

Cell In[13], line 6, in Thing.my(self)
      5 def my(self) -> int:
----> 6     return hello

NameError: name 'hello' is not defined
```

I expect this will be replaced with typing.Self from python 3.11?

castep_outputs/utilities/filewrapper.py
```diff
@@ -278,8 +280,7 @@ def remove_bounds(self, /, fore: int = 1, back: int = 2) -> None:
             Whether to strip trailing line from data.
         """
         self._lineno += fore
-        self._data = self._data[fore:
-                                -back if back else None]
+        self._data = self._data[fore : -back if back else None]
```
ajjackson, line 300: This can be `-back or None`, but I'm not sure if it _should_ be.

castep_outputs/utilities/utility.py
```diff
@@ -128,7 +161,18 @@ def atreg_to_index(dict_in: dict[str, str] | re.Match, *, clear: bool = True) ->
...
+
+@overload
+def normalise(obj: T, mapping: NormDict) -> T: ...
+@overload
+def normalise(obj: In, mapping: NormDict) -> Out: ...
+@overload
+def normalise(obj: Iterable[In | T], mapping: NormDict) -> tuple[Out | T, ...]: ...
+@overload
+def normalise(obj: Mapping[K, In | T], mapping: NormDict) -> dict[K, Out | T]: ...
+def normalise(obj, mapping):
```
ajjackson, line 175: Some pretty epic annotation here to make such a flexible function work! Hopefully it pays off in API simplicity elsewhere.

castep_outputs/utilities/utility.py
ajjackson: Seems best to make this consistent with type signature.
```suggestion
            as a set of strings.
```

ajjackson (changes requested):
This looks good! How grueling was it to get here?

Just a couple of suggestions, one of which is questionable.

(And some merge conflicts to resolve.)

castep_outputs/utilities/filewrapper.py
oerc0122: Strange, I thought I learned this from you...? 

Yeah, it will become `typing.Self`.

castep_outputs/utilities/filewrapper.py
oerc0122: I agree. Although it does mean the `back==0 or back is None` case more explicit if I actually express that.

castep_outputs/utilities/utility.py
oerc0122: Yeah, it's a recursive call to normalise a bunch of objects (as you can see). The 3.12+ signature will look nicer (since the names will be local)

castep_outputs/utilities/filewrapper.py
ajjackson: > Strange, I thought I learned this from you...?

Entirely possible, it does seem like something one might forget until it next comes up!

ajjackson (approved):

PR #248, opened 2026-04-08
Rework binary reader to be more descriptive

oerc0122:
Make binary reader a class with descriptive methods for complicated data processing.

### DIFF
--- a/castep_outputs/bin_parsers/fortran_bin_parser.py
+++ b/castep_outputs/bin_parsers/fortran_bin_parser.py
@@ -1,12 +1,16 @@
 """General parser for the Fortran Unformatted file format."""
-from collections.abc import Generator
+
+from collections.abc import Iterable, Iterator, Mapping
 from os import SEEK_CUR
-from typing import BinaryIO
+from typing import BinaryIO, TypeVar
 
-FortranBinaryReader = Generator[bytes, int, None]
+from castep_outputs.utilities.utility import to_type
+
+T = TypeVar("T")
+K = TypeVar("K")
 
 
-def binary_file_reader(file: BinaryIO) -> FortranBinaryReader:
+class FortranBinaryReader:
     """Yield the elements of a Fortran unformatted file.
 
     Parameters
@@ -18,11 +22,6 @@ def binary_file_reader(file: BinaryIO) -> FortranBinaryReader:
     ------
     bytes
         Binary data record from Fortran file.
-
-    Receives
-    --------
-    int | bool
-        Skip/rewind amount (``True`` == ``-1``).
 
     Notes
     -----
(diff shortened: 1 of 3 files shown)


castep_outputs/bin_parsers/fortran_bin_parser.py
```diff
@@ -40,22 +39,123 @@ def binary_file_reader(file: BinaryIO) -> FortranBinaryReader:
...
+        dtypes : dict[K, type[T]]
+            Dictionary mapping key names to dtype to read.
+
+        Returns
+        -------
+        dict[K, T]
+            Loaded data.
+        """
+        accum: dict[K, T] = {}
+        for (key, typ), datum in zip(dtypes.items(), self, strict=False):
```
ajjackson, line 158: At last, an elegant use for `zip(... strict=False)`!

test/test_binary_reader.py
```diff
@@ -1,24 +1,59 @@
...
+@pytest.mark.parametrize("fake_file", [(1,)], indirect=True)
+def test_reader_get(fake_file):
+    reader = FortranBinaryReader(fake_file)
+    assert reader.get(int) == 1
+
+
+@pytest.mark.parametrize("fake_file", [(1, 2, 3, 2)], indirect=True)
+def test_reader_get_iter(fake_file):
+    reader = FortranBinaryReader(fake_file)
+    assert list(reader.get_dtype_iter((int, int, int, int))) == [1, 2, 3, 2]
```
ajjackson, line 26: Very nice use of `indirect` magic. Stealing this.

test/test_binary_reader.py
```diff
@@ -29,36 +64,44 @@ def test_actual_read():
...
-            assert to_type(datum, typ) == expected
+        reader = FortranBinaryReader(file)
+        for (typ, expected) in dtypes:
+            assert reader.get(typ) == expected
 
-        for (n, datum) in enumerate(reader):
+        for n, datum in enumerate(reader):
             x, y = divmod(n, 16)
             ind = to_type(datum[:8], int)
+            print(ind, (x + 1, y + 1))
```
ajjackson:
```suggestion
```

ajjackson (commented):
This looks like a nice delegation of responsibility/knowledge to an appropriate class.

It does seems like a lot more code added than removed to achieve it, though. Presumably that's because not all the features of the new class are being used yet. Where do you anticipate that they will be?

oerc0122:
> This looks like a nice delegation of responsibility/knowledge to an appropriate class.
> 
> It does seems like a lot more code added than removed to achieve it, though. Presumably that's because not all the features of the new class are being used yet. Where do you anticipate that they will be?

One of the big delays in writing the new binary parsers is working out the structure of the data from the source, being able to copy and paste (with minimal modification) some of these things into the readers and get a rough outline of how the data look is one of the things that will help with development if not implementation of these things.

ajjackson (approved):
I see! Then this is not just a refactor but also providing features to support further development.

It looks fine, is tested, and is part of a wider plan. Sounds like it should be merged soon 😄

PR #249, opened 2026-04-14
Fix some get_generated_files and docs issues

oerc0122:
- Add some documentation on tools
- Fix issue with main parser docs given rename
- Fix issue with singledispatch in get_generated_files
- Fix missing `write_orbitals` issue.

### DIFF
--- a/docs/source/tools.ipynb
+++ b/docs/source/tools.ipynb
@@ -9,8 +9,8 @@
     "\n",
     "## get_generated_files\n",
     "\n",
-    "`get_generated_files` is a function which can predict the files which will be produced by a CASTEP\n",
-    "run. It can accept either parsed (via :func:`parse_cell_param_file`) ``.cell`` / ``.param`` files or\n",
+    "`get_generated_files` is a function which can predict the files that will be produced by a CASTEP\n",
+    "run. It can accept either parsed (via `parse_cell_param_file`) `.cell` / `.param` files or\n",
     "paths or a seedname."
    ]
   },
@@ -64,7 +64,7 @@
    "source": [
     "## md_geom_parser\n",
     "\n",
-    "The ``md_geom_parser`` is a class which provides an iterative interface for loading md/geom data\n",
+    "The `MDGeomParser` is a class which provides an iterative interface for loading md/geom data\n",
     "without loading the whole file simultaneously. It supports efficient [Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence)\n",
     "style access to the data, returning parsed frame data or lists thereof."
    ]
--- a/docs/source/tools.ipynb
+++ b/docs/source/tools.ipynb
@@ -29,7 +29,7 @@
     "param_path = example_path.with_suffix(\".param\")\n",
     "\n",
     "# Automatically picks up cell/param files if present.\n",
-    "get_generated_files(example_path)"
+    "get_generated_files(seedname=example_path)"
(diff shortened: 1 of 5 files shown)


docs/source/tools.ipynb
```diff
@@ -0,0 +1,151 @@
...
+   "cell_type": "markdown",
+   "id": "a98ef7ef-9e8d-44d4-bd4c-a18e71e5ef99",
+   "metadata": {},
+   "source": [
+    "# Tools\n",
+    "\n",
+    "## get_generated_files\n",
+    "\n",
+    "`get_generated_files` is a function which can predict the files which will be produced by a CASTEP\n",
+    "run. It can accept either parsed (via :func:`parse_cell_param_file`) ``.cell`` / ``.param`` files or\n",
```
ajjackson:
```suggestion
    "`get_generated_files` is a function which can predict the files that will be produced by a CASTEP\n",
    "run. It can accept either parsed (via `parse_cell_param_file`) `.cell` / `.param` files or\n",
```

ajjackson: I think .ipynb only supports markdown

docs/source/tools.ipynb
```diff
@@ -0,0 +1,151 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "aca441d6-6ac7-4646-b90f-f360ddc94bfd",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pathlib import Path\n",
+    "from castep_outputs.tools.get_generated_files import get_generated_files\n",
+    "\n",
+    "example_path = Path.cwd() / \"example_data\" / \"pair-pot-lj\"\n",
```
ajjackson, line 27: Initially I read this as meaning "pair-pot-lj" would be a directory. Not sure about most elegant way to improve that, it's fine really.

docs/source/tools.ipynb
```diff
@@ -0,0 +1,151 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5440c208-165f-4653-958d-7434e0904019",
+   "metadata": {},
+   "source": [
+    "## md_geom_parser\n",
+    "\n",
+    "The ``md_geom_parser`` is a class which provides an iterative interface for loading md/geom data\n",
```
ajjackson:
```suggestion
    "The `MDGeomParser` is a class which provides an iterative interface for loading md/geom data\n",
```

ajjackson (commented):
All looks reasonable! A couple of comments here on the .ipynb

docs/source/tools.ipynb
oerc0122: Thanks for spotting that! I originally wrote this as a `.rst` file, then copied and pasted it into a notebook because I thought a run might be more useful.

docs/source/tools.ipynb
oerc0122: Putting `seedname=` explcitly into the `get_generated_files` call? calling it `seedname`?

ajjackson: Using `seedname=` seems a good compromise because it is _sort of_ a path and _sort of_ a seedname.

In fact it is teaching that in this code `seedname` is allowed to include a path prefix.

ajjackson (changes requested):
I think that would work, good idea 😄

ajjackson (approved):

PR #252, opened 2026-04-17
Fix cell/param parser and get_generated files

oerc0122:
Apply changes and fixes such that all cell/param in the castep test suite pass.

### DIFF
--- a/castep_outputs/parsers/cell_param_file_parser.py
+++ b/castep_outputs/parsers/cell_param_file_parser.py
@@ -10,17 +10,16 @@ from functools import partial
 from typing import Any, Literal, TextIO, TypedDict
 
 import castep_outputs.utilities.castep_res as REs
-
-from ..utilities.constants import SHELLS
-from ..utilities.datatypes import (
+from castep_outputs.utilities.constants import SHELLS
+from castep_outputs.utilities.datatypes import (
     AtomIndex,
     MaybeSequence,
     PSPotStrInfo,
     ThreeByThreeMatrix,
     ThreeVector,
 )
-from ..utilities.filewrapper import Block
-from ..utilities.utility import (
+from castep_outputs.utilities.filewrapper import Block
+from castep_outputs.utilities.utility import (
     atreg_to_index,
     determine_type,
     file_or_path,
@@ -76,7 +75,8 @@ DevelElem = MaybeSequence[str | float | dict[str, str | float]]
 DevelBlock = dict[str, DevelElem | dict[str, DevelElem]]
 HubbardU = dict[str | AtomIndex, str | dict[str, float]]
 CellParamData = dict[
-    str, str | float | tuple[float, str] | dict[str, Any] | HubbardU | DevelBlock | XCDef,
+    str,
+    str | float | tuple[float, str] | dict[str, Any] | HubbardU | DevelBlock | XCDef,
 ]
 GeneralBlock = dict[
     str,
@@ -147,14 +147,18 @@ def parse_cell_param_file(cell_param_file: TextIO) -> list[CellParamData]:
             curr[block_title] = _PARSERS.get(block_title, _parse_general)(block)
 
         elif match := REs.PARAM_VALUE_RE.match(line):
+
             key, val, unit = match.group("key", "val", "unit")
             key = key.strip().lower()
             logger("Found param %s", key)
 
-            if " " in val.strip():
+            if determine_type(val) is str:
+                val = val.strip()
+            elif " " in val.strip():
                 val = to_type(val.split(), determine_type(val))
             else:
                 val = to_type(val, determine_type(val))
+
             if unit:
                 unit = unit.strip()
                 curr[key] = (val, unit)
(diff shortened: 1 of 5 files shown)


castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -76,7 +75,8 @@ class XCDef(TypedDict):
 DevelBlock = dict[str, DevelElem | dict[str, DevelElem]]
 HubbardU = dict[str | AtomIndex, str | dict[str, float]]
 CellParamData = dict[
-    str, str | float | tuple[float, str] | dict[str, Any] | HubbardU | DevelBlock | XCDef,
+    str,
+    str | float | tuple[float, str] | dict[str, Any] | HubbardU | DevelBlock | XCDef,
```
ajjackson, line 79: It looks like `DevelBlock` and `XCDef` are included in `dict[str, Any]`. Is there an aspiration here to eventually ditch `dict[str, Any]` or are there features that will always require it?

castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -151,10 +151,14 @@ def parse_cell_param_file(cell_param_file: TextIO) -> list[CellParamData]:
             key = key.strip().lower()
             logger("Found param %s", key)
 
-            if " " in val.strip():
+            if determine_type(val) is str:
+                val = (val + " " + (unit or "")).strip()
```
ajjackson: This feature looks a bit weird to me, because units are usually returned as a tuple element. Is it supposed to fix cases where the last word of phrase was incorrectly plucked away as a "unit"? May be worth a comment if so.

castep_outputs/tools/get_generated_files.py
```diff
@@ -28,7 +31,7 @@ class UCEnum(Enum):
     def _missing_(cls, value: Any) -> Self | None:
         if not isinstance(value, str):
             return None
-        value = value.upper()
+        value = value.translate(TRANS_TABLE)
```
ajjackson, line 34: What case is this needed for? Slightly scary thing to see, makes me wonder what other code can have trouble!

castep_outputs/parsers/cell_param_file_parser.py
oerc0122: The issue is that there are other less-specific `dict`s that aren't defined yter and until I clarify those I think it's good to keep note of the fact there are specific ones until we can worm out exactly. 

Once the castep cell/param yaml dumper is released it'll be possible to generate actual TypedDict specs for cell/param file parsing, which would be handy.

castep_outputs/parsers/cell_param_file_parser.py
oerc0122: Yes, it is bizarre because I wrote it in a hurry. It definitely need cleanup. I honestly thought I'd opened this as a draft.

The issue is that `unit` can be `None` or the last word of e.g. `single point`.

castep_outputs/tools/get_generated_files.py
oerc0122: This is because all of the keys of the `Enum` are upper-case, no spaces and no non-underscore punctuation. However, one of the keys castep allows is `PHONON+EFIELD` and there's no trivial way to filter that. It would become something like:
```python
value = "".join(x for x in value if x.isalpha() or x == "_").upper()
```
Which may or may not be clearer (though the `_` check could strictly be removed)?

ajjackson: How about a `.replace("+", "_")`?

oerc0122:
```python
value = "".join(x.upper().replace("+", "_").split())
```
Is still much less readable to me than the translation table. And if there's other punctuation that could go in there e.g. "-", it now becomes:
```python
value = "".join(x.upper().replace("+", "_").replace("-", "_").split())
```
Rather than:
```python
TRANS_TABLE = str.maketrans(lower, upper, "+ ")
```
->
```python
TRANS_TABLE = str.maketrans(lower+"-", upper+"_", "+ ")
```
or
```python
TRANS_TABLE = str.maketrans(lower, upper, "-+ ")
```

ajjackson: Ah right! Yeah, if we are only removing other characters then that does seem to be the exact tool for the job.

castep_outputs/parsers/cell_param_file_parser.py
ajjackson: Ok. Maybe we can have a milestoned issue or something tracking the aspiration to cover other dicts and ditch the `dict[str,Any]`.

castep_outputs/parsers/cell_param_file_parser.py
ajjackson: Maybe this block should be extracted to a method while making it clearer that the last group isn't always a unit:

```python
def _process_param_entry(key: str, val: str, suffix: str | None) -> tuple[str, VALTYPES, str | None]:
    key, val, suffix = (item.strip() for item in key, val, suffix if item)
    key = key.lower()

    val_type = determine_type(val)
    if val_type is str:
        # Suffix is part of string
        return key, f"{val} {suffix}", None

    if " " in val.strip():
        val = val.split()

    # suffix is unit
    return key, to_type(val, val_type), suffix
...

elif match := REs.PARAM_VALUE_RE.match(line):
    key, val, unit = _process_param_entry(*match.group("key", "val", "suffix"))
    logger("Found param %s", key)
    curr[key] = (val, unit) if unit else val

```

oerc0122: There we go, much cleaner solution.

ajjackson:
```python
           if determine_type(val) is str:
                val = val.strip()
            elif " " in val.strip():
                val = to_type(val.split(), determine_type(val))
            else:
                val = to_type(val, determine_type(val))
```
Not very DRY but it does make the three possible cases nice and clear. This'll do!

ajjackson (approved):

PR #257, opened 2026-04-22
Fix castep_outputs to parse all benchmark.out

### LINKED ISSUE #237: pspot string parser
Hi! 

I have a .castep file with this pspot string:

"3|2.4|2.4|1.0|8|9|10|40:41:32U2U2(qc=5.5)"

that `_parse_pspot_string` doesn't like

```
_parse_pspot_string("3|2.4|2.4|1.0|8|9|10|40:41:32U2U2(qc=5.5)")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 _parse_pspot_string("3|2.4|2.4|1.0|8|9|10|40:41:32U2U2(qc=5.5)")

File ~/coding/castep_outputs/castep_outputs/parsers/cell_param_file_parser.py:176, in _parse_pspot_string(string, debug)
    174 def _parse_pspot_string(string: str, *, debug: bool = False) -> PSPotStrInfo:
    175     if not (match := REs.PSPOT_RE.search(string)):
--> 176         raise ValueError(f"Attempt to parse {string} as PSPot failed")
    178     pspot = match.groupdict()
    179     projectors = []

ValueError: Attempt to parse 3|2.4|2.4|1.0|8|9|10|40:41:32U2U2(qc=5.5) as PSPot failed

```

Changing line 314 in castep_res.py to

`PROJ_TYPES = r"[UNGHLP\d]"`

seems to fix the issue for me.

### LINKED ISSUE #256: External files not parsed
castep_file_parser is not picking up external files.

oerc0122:
Fixes #256 
Fixes #237 

### `cell/param`
- Rewrite PSPot string parser
- Attach units to SEDC
### `castep`
- Fix some edge cases in MD/geom sections (final steps without initial)
- Fix issue with finalisation -> <BEGIN ...>
- Fix issue with "generated on-the-fly" pspot in `.castep`
- Fix some issues with messages in phonon `char_table`
### `fmt`
- Use `next` over `readline` to allow `Blocks`
### `gen_data`
- `Make all` by default


test/test_castep_parser.py
ajjackson, line 1: You have a failing test here because a `defaultdict` is compared with a regular `dict`. That's easily covered by casting to `dict` in the test before comparing... but maybe it should be done in the actual function? What does the function signature promise?

oerc0122: That's actually not the origin of the error, the error was that the dicts were different, pytest only checks the contents of dict-likes. 

In the ideal case, the function signature promise is a `TypedDict` which is neither a `dict` nor a `defaultdict`, but `defaultdict` is a strict subclass of `dict` and thus meets the requirements of typing.

```
>>> from collections import defaultdict
>>> issubclass(defaultdict, dict)
True
```

ajjackson: Ah! Didn't realise it was a subclass but actually that makes sense 🤦

oerc0122: Yeah, it just has a factory `_missing_` implementation.

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -1036,14 +1043,20 @@ def parse_castep_file(castep_file_in: TextIO,
...
+            castep_file,
+            "Final Configuration",
+            rf"\s*{REs.MINIMISERS_RE}\s*: Final",
+        ):
             if Filters.GEOM_OPT not in to_parse and Filters.FINAL_CONFIG not in to_parse:
                 continue
 
             logger("Found final geom configuration")
 
+            if "geom_opt" not in curr_run:
```
ajjackson: How does this situation arise? Is it
- some verbosity levels (or other parameter sets) don't give any geom opt information before this during task=GeometryOptimization, or
- this catches "Final Configuration" from a task that isn't GeometryOptimization, such as TransitionStateSearch?

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -1036,14 +1043,20 @@ def parse_castep_file(castep_file_in: TextIO,
...
+                curr_run["geom_opt"] = defaultdict(list)
```
ajjackson: Also, could/should we eliminate the `if` by using `setdefault`?

castep_outputs/parsers/castep_file_parser.py
```diff
@@ -2369,19 +2390,33 @@ def _process_phonon(block: Block, logger: Logger) -> list[QData]:
...
+            char_table = cast("Iterator[str]", char_table)
+
             next(char_table)
             char: list[CharTable] = []
+
             for char_line in char_table:
                 if re.search(r"[-=]{4,}", char_line):
                     break
+                if char_line.startswith(" Sum over"):
+                    break  # TODO: Implement parser  # noqa: FIX002
```
ajjackson: If you are using a TODO workflow in this project then disable FIX002 globally.
(Or disable it for this file in pyproject.toml if you aspire to FIX002 but can't right now.)

If using TODO systematically perhaps it would be nice to have an action that counts them and displays it somewhere ;-) Or a release pipeline that refuses to release if TODO remain in the codebase?

castep_outputs/parsers/castep_file_parser.py
oerc0122: This may, in fact, have been a false positive from one of the other fixes not being fixed yet.

castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -169,47 +169,82 @@ def parse_cell_param_file(cell_param_file: TextIO) -> list[CellParamData]:
...
-    pspot = match.groupdict()
-    projectors = []
-
-    for proj in pspot["proj"].split(":"):
-        if match := REs.PSPOT_PROJ_RE.match(proj):
-            pdict = dict(zip(REs.PSPOT_PROJ_GROUPS, match.groups(), strict=True))
+    logging.debug("%s", string)
+
+    pspot: dict[str, Any] = {
+        "print": "[" in string,
```
ajjackson, line 267: Not sure how to interpret this. Are we relying on brackets after a `#` being stripped off? Otherwise, all the OTFPP strings I've seen have brackets at the end. Chris Pickard's write-up says "Removing the # activates full testing" but I don't really know what that means. Presumably it means that if you include a modification to default configuration that will always go though "full testing" as well.

castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -169,47 +169,82 @@ def parse_cell_param_file(cell_param_file: TextIO) -> list[CellParamData]:
...
+        "poly_fit": "-" in string[:string.find("|")],
```
ajjackson: Can you give an example somewhere of what this is looking for? I can't make sense of it from https://www.tcm.phy.cam.ac.uk/castep/otfg.pdf or https://castep-docs.github.io/castep-docs/documentation/Pseudopotentials/otfg_string , are there any docs for this feature?

castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -169,47 +169,82 @@ def parse_cell_param_file(cell_param_file: TextIO) -> list[CellParamData]:
...
+                raise ValueError(f"Attempt to parse {string!r} as PSPot failed.")
+
+            pspot[label] = match[1].split(",")
+            string = string[: match.start()] + string[match.end() :]
         else:
-            raise ValueError("Invalid PSPot string")
+            pspot[label] = None
+
+    if not (match := REs.PSPOT_INFO_RE.match(string)):
+        logging.error("Failed parsing main.")
```
ajjackson: Will it be obvious in the log that "main" refers to part of the OTFG string?

castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -169,47 +169,82 @@ def parse_cell_param_file(cell_param_file: TextIO) -> list[CellParamData]:
...
+        raise ValueError(f"Attempt to parse {string!r} as PSPot failed.")
 
-        pdict["shell"] = SHELLS[int(pdict["shell"])]
+    logging.debug("Main: %s", match.groupdict())
 
-        if not pdict["type"]:
-            pdict["type"] = None
+    pspot.update(match.groupdict())
+    string = string[: match.start()] + string[match.end() :]
+    pspot["proj"] = string
```
ajjackson: Does "proj" mean something other than "projectors"? Seems odd that `pspot` has both of these as keys.

castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -169,47 +169,82 @@ def parse_cell_param_file(cell_param_file: TextIO) -> list[CellParamData]:
...
 
-        for prop in ("beta_delta", "de"):
-            if not pdict[prop]:
-                del pdict[prop]
+    for projectors in string.split(":"):
+        logging.debug("Full projector: %s", projectors)
+        proj = {
```
ajjackson: I'm finding this too hard to review: it's exhausting to build a mental model of this code and follow what is going on. "proj" appears again as a variable name here – not to be confused with `projectors`, `proj_data`, `pspot["proj"]`, `pspot["projectors"]` or `proj["projectors"]` – and is reassigned on each iteration of this loop scope and then one of its elements is updated inside the next loop before the thing is appended back to the parent. So, if I have it right: `pspot["projectors"]` contains a list of `proj` which are dicts containing their own "projecters" item which is a list of `proj_data` dicts from a regex. And I need to know all that to understand this part of the code.

I get that this reflects the underlying complexity of the data, but perhaps there is another way?

ajjackson: (Apologies if tone seems grouchy there, I love what you're doing in general 🤗 )

castep_outputs/parsers/castep_file_parser.py
oerc0122: Ech, it's more that I know this is a hack, it's a 3 second fix. They're tiny pieces of cleanup, I just think this PR is already doing a bit too much and these are far less important than getting the `benchmark.out`s parsing for the test harness.

ajjackson: Best do it as `[tool.ruff.lint.extend-per-file-ignores]` then, that's both neater and more visible.

castep_outputs/parsers/cell_param_file_parser.py
oerc0122: `proj` is the original string from which the `projectors` are derived. Not sure how to put it.

castep_outputs/parsers/cell_param_file_parser.py
oerc0122: `#` is a comment character, any `[]` activates testing. `#` is not really a special character it's just, as you say, stripping off the bracket.

castep_outputs/parsers/cell_param_file_parser.py
ajjackson: From discussion:
- Poly fit is indicated by surrounding the first token with `-`
- No, this doesn't seem to be documented anywhere

ajjackson: Marking "resolved" as there's not a lot to be done, and this is now more bearable in the generally-improved state of things 😄

castep_outputs/parsers/cell_param_file_parser.py
ajjackson: Updated to "beta_function_string"

castep_outputs/parsers/cell_param_file_parser.py
```diff
@@ -168,19 +170,100 @@ def parse_cell_param_file(cell_param_file: TextIO) -> list[CellParamData]:
...
-        "poly_fit": "-" in string[:string.find("|")],
+        "poly_fit": "-" in string[: string.find("|")],
         "beta_functions": [],
         "string": string,
     }
 
-    # Remove polyfit
-    string = re.sub(r"-?([^|]+)-?\|", r"\g<1>|", string, count=1)
+    # Remove polyfit `-` from string
+    string = re.sub(r"([^|]*)-([^|]*)\|", r"\g<1>\g<2>|", string, count=1)
```
ajjackson: Is this not going to strip the `-` from a negative <dE_use> parameter? https://castep-docs.github.io/castep-docs/documentation/Pseudopotentials/otfg_string/

(I still don't know what a string with polyfit looks like or does, is it worth looking in the source code?)

castep_outputs/parsers/cell_param_file_parser.py
ajjackson: Extracting two functions has made a huge improvement to readability 👍

castep_outputs/parsers/cell_param_file_parser.py
oerc0122: Ah! In the rework I missed one key thing
```suggestion
    string = re.sub(r"^([^|]*)-([^|]*)\|", r"\g<1>\g<2>|", string, count=1)
```

castep_outputs/parsers/cell_param_file_parser.py
oerc0122: > From discussion:
> 
> * Poly fit is indicated by surrounding the first token with `-`
> * No, this doesn't seem to be documented anywhere

Not surrounding, just a `-` appearing anywhere in the first block

test/test_cell_param_parser.py
```diff
@@ -550,14 +551,16 @@ def test_xc_def(self):
     "2|1.8|1.8|1.3|2|3|4|30:31:32LGG(qc=4)",
     "2|1.8|3.675|5.512|7.35|30UU:31UU:32LGG[]",
     "2|1.8|3.675|5.512|7.35|30UU:31UU:32LGG{1s1}[]",
-    "3|2.4|2.4|1.0|8|9|10|40:41:32U2U2(qc=5.5)"
+    "3|2.4|2.4|1.0|8|9|10|40:41:32U2U2(qc=5.5)",
+    "3|1.7|1.7|0.9|14|16|18|40U:41U:42:50:51{5s0.5,5p0.5}(qc=6)[]",
+    "3|1.6|16|16|19|40U:50:41:42(qc=7){5s0.5}",
+    "2|2|2|1.4|8|11|11|50N:60NH:51UU:52LGG:43U1.6+0U1.6+0.1{5d0.9,4f0.1}(qc=6)"
```
ajjackson, line 566: Would this be a good time to generate/sanity-check some outputs of the function?
Preferably including a polyfit-enabled case and a negative-energy case to make sure that is all picked up / resolved as expected.

ajjackson: Perhaps that can be an Issue beyond scope of the PR. Bonus points if it's done while sitting with a senior CASTEP developer who can check the answers.

oerc0122: Yeah, it's part of the reason why I hadn't had any of the tests until now. Need to make sure they're reasonable.

ajjackson: Ok, let's not have that block the PR.

ajjackson (approved):

PR #260, opened 2026-04-27
Fix issue with reading external files

oerc0122:
- Old end_re didn't allow `<begin XXX>` blocks.
  - If present back up and process 
- `readline` is not present on `FileWrapper`s, `next` is universal.

### DIFF
--- a/castep_outputs/utilities/utility.py
+++ b/castep_outputs/utilities/utility.py
@@ -226,7 +226,7 @@ def normalise(obj, mapping):
     Returns
     -------
     Any
-        Normmalised data.
+        Normalised data.
     """
     if isinstance(obj, (tuple, list)):
         obj = tuple(normalise(v, mapping) for v in obj)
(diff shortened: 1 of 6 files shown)


castep_outputs/utilities/utility.py
```diff
@@ -326,7 +326,8 @@ def flatten_dict(
     """
     items: list[tuple[Any, Any]] = []
     for key, value in dictionary.items():
-        new_key = str(parent_key) + separator + key if parent_key else key
+        # Temporary hack because `None` used as key in species_pot
+        new_key = str(parent_key) + separator + str(key) if parent_key else key
```
ajjackson:
```suggestion
        new_key = f"{parent_key}{separator}{key}" if parent_key else key
```

ajjackson:
Is there test coverage for the `<BEGIN ...>` case? I see there is a parsing block for it further down. Or perhaps there are other places it can appear, so that block is already covered?

ajjackson (approved):
Ok, looks ready to go!

PR #262, opened 2026-05-01
Fix to_type to more strict definition.

### LINKED ISSUE #261: Change to_type functionality
Ok, I'd be happy to push back this discussion to another issue/time.

But it sounds like it would make sense to initially treat the binary data as _always_ returning a tuple, and explicitly cast some fields to scalar. i.e.

```python
    def get_dtype_cycle(
        self,
        dtypes: tuple[type[T], ...],
        *,
        n: int | None = None,
    ) -> Iterator[tuple[tuple[T, ...], ...]]:
...

    for data in reader.get_dtype_cycle(dtypes, n=total):
        ns = as_scalar(data.pop("ns"))
        nk = as_scalar(data.pop("nk"))
        data["n_bands"] = as_scalar(data["n_bands"])
        data["min_band"] = as_scalar(data["min_band"])

        accum["kpoints"][ns, nk] = data
```

where `as_scalar()` is something like [more_itertools.only](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.only)

(I appreciate that sketch isn't very DRY and can be improved further...)

_Originally posted by @ajjackson in https://github.com/oerc0122/castep_outputs/pull/246#discussion_r3162778242_

oerc0122:
- Create new `type_conv` module for type conversion from `utility`
- Split `to_type` into `to_type` (performing similar to original functionality on `str` and `parse_bytes`
- `parse_bytes` handles bytes and takes either a type (e.g. `int`) or a tuple (e.g. `(int, ...)` [literal ellipsis]) returning either a scalar (and failing if non-scalar) or sequence respectively.
- Remove indirect references to imports (`..`)

Fixes #261


castep_outputs/utilities/utility.py
```diff
@@ -495,7 +496,7 @@ def determine_type(data: str) -> type:
     return str
 
 
-def parse_int_or_float(numbers: Iterable[str]) -> int | float:
+def parse_int_or_float(numbers: Iterable[str]) -> tuple[int | float, ...]:
```
ajjackson: This is just a correction to the annotation for existing behaviour, right?

castep_outputs/utilities/utility.py
```diff
@@ -672,73 +670,117 @@ def _parse_complex_bytes(val: bytes) -> complex | Sequence[complex]:
...
+        case (type() as type_, EllipsisType()):
+            parser = parsers[type_]
+            out = parser(data_in)
+        case type() as type_:
+            parser = parsers[type_]
+            out = get_only(parser(data_in))
+        case _:
+            raise TypeError(f"Cannot handle converting to {typ!r}.")
+
+    return out
```
ajjackson: Perhaps I've misread this somehow, but it looks to me that the setup here is
```python
to_type("1", complex) -> 1+0j
to_type("1", (complex, ...)) -> (1+0j,)
to_type(["1", "2"], complex) -> (1+0j, 2+0j)
to_type(["1", "2"], (complex, ...)) -> ((1+0j,), (2+0j,))
```
so the `(type, ...)` notation isn't actually useful for creating collections of multiple data, and `typ` won't necessarily match the return type (because the data might be a sequence.).

Does this logic come together somewhere else? Otherwise I would expect

```python
to_type("1", complex) -> 1+0j
to_type("1", (complex, ...)) -> TypeError
to_type(["1", "2"], complex) -> TypeError
to_type(["1", "2"], (complex, ...)) -> (1+0j, 2+0j,)
```

i.e.

```python
@overload
def to_type(data_in: T, typ: Any) -> T: ...
@overload
def to_type(data_in: str | bytes, typ: type[T]) -> T: ...
@overload
def to_type(data_in: Iterable[str | bytes], typ: ToTypeTuple) -> tuple[T, ...]: ...
```

castep_outputs/utilities/utility.py
oerc0122: No, this is a re-engineering so that all the parsing functions return a `tuple[T]` for `to_type` to destructure if requested.

oerc0122: It's also actually wrong.

oerc0122:
```suggestion
def parse_int_or_float(numbers: Iterable[str]) -> int | float:
```

castep_outputs/utilities/utility.py
oerc0122: The real meaning is that (using pseudo-bytes here):
```python
to_type(b"1 2 3", int) -> TypeError
to_type(b"1 2 3", (int, ...)) -> (1, 2, 3)
to_type(b"1", int) -> 1
to_type(b"1", (int, ...)) -> (1,)
```
It is generally the case that strings are split beforehand while bytes aren't really parsed as "binary strings". Maybe this is overloading functionality we don't want. 
In actual bytes this might be:
```python
to_type(b"\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00\x03", (int, ...)) -> (1, 2, 3)
```

test/test_castep_parser.py
```diff
@@ -3185,25 +3185,25 @@ def test_geom_output(self):
...
                                         "value": 0.4266993,
                                     },
-                                    "dr_max": {
+                                    "smax": {
                                         "converged": False,
-                                        "tolerance": 0.001,
-                                        "value": 0.0,
+                                        "tolerance": 0.25,
+                                        "value": 12.78436,
                                     },
```
ajjackson, line 3207: This reordering of reference dicts looks unrelated/spurious?

ajjackson (approved):
This looks really nice to me! With this separation of concerns between `bytes` and `str` handling each implementation is simple to understand. The calls are also simple to understand, without getting significantly more verbose or inconvenient.

test/test_castep_parser.py
oerc0122: Yeah, that's the actual order that falls out of pprint from when I compare things, I can revert it here, or just ignore it.

ajjackson: Should revert really for cleaner history, but the numbers haven't changed so I'm not worried.

oerc0122: It just means I have to re-sort when I want to manually compare the values if they change. Going to keep this and get this merged so all my other PRs break.

PR #264, opened 2026-05-05
Clean up and unify parsing md/geom header

oerc0122:
- Unify header parsing between lazy and main md/geom parsers.
- Add test for parsing header.

### DIFF
--- a/castep_outputs/parsers/md_geom_file_parser.py
+++ b/castep_outputs/parsers/md_geom_file_parser.py
@@ -8,7 +8,7 @@ from typing import TextIO, TypedDict
 from castep_outputs.utilities.castep_res import ATOMIC_DATA_TAG, TAG_RE, get_numbers
 from castep_outputs.utilities.constants import FST_D, TAG_ALIASES
 from castep_outputs.utilities.datatypes import AtomIndex, ThreeByThreeMatrix, ThreeVector
-from castep_outputs.utilities.filewrapper import Block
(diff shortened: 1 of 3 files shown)


castep_outputs/parsers/md_geom_file_parser.py
```diff
@@ -108,6 +108,37 @@ def parse_md_geom_frame(block: Block) -> MDGeomTimestepInfo:
...
+        Comment from header.
+    """
+    logger = log_factory(md_geom_file)
+
+    line = ""
+    for line in md_geom_file:
+        if line.strip():
+            break
+
+    if block := Block.from_re(line, md_geom_file, "BEGIN header", "END header"):
```
ajjackson, line 131: Not really specific to this PR, but:

These `.from_re` calls would be a bit easier to understand if the `from_re` docstring explained what assumptions it makes about the relationship between `init_line` and `in_file`. I think the expectation is that `in_file` is in a state to begin iterating lines directly _after_ `init_line`, right?

ajjackson (approved):
Looks clean and with a nice test update 👍

castep_outputs/parsers/md_geom_file_parser.py
oerc0122: `init_line` is checked against `start`, then if there's a match, it iterates `in_file` until `end` is matched `n_end` times.

PR #266, opened 2026-05-07
Fix typing in FortranBinaryReader

oerc0122:


ajjackson (approved):
Looks correct to me 😄

PR #267, opened 2026-05-07
Clean up typing docs

oerc0122:
Built on fix-typing

Refactors typing to use native sphinx typehints and display properly.

### DIFF
--- a/castep_outputs/parsers/md_geom_file_parser.py
+++ b/castep_outputs/parsers/md_geom_file_parser.py
@@ -41,7 +41,7 @@ class MDGeomTimestepInfo(TypedDict, total=False):
 
     #: Elapsed MD Time.
     time: float
-    #: Current energies: total, potential, kinetic.
+    #: Current energies꞉ total, potential, kinetic.
     energy: tuple[float, float, float]
     #: Instantaneous temperature.
     temperature: tuple[float]
@@ -81,7 +81,7 @@ def parse_md_geom_frame(block: Block) -> MDGeomTimestepInfo:
 
     Returns
     -------
-    MDGeomTimestepInfo
+    :
         Parsed frame of data.
     """
     curr: MDGeomTimestepInfo = defaultdict(list)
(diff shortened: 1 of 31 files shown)


castep_outputs/parsers/md_geom_file_parser.py
```diff
@@ -41,7 +41,7 @@ class MDGeomTimestepInfo(TypedDict, total=False):
 
     #: Elapsed MD Time.
     time: float
-    #: Current energies: total, potential, kinetic.
+    #: Current energies; total, potential, kinetic.
```
ajjackson: This should be a colon, though; it introduces a list of items.

pyproject.toml
```diff
@@ -33,13 +33,13 @@ ruamel = ["ruamel.yaml>=0.17.22,<0.19"]
 yaml = ["pyYAML>=3.13"]
 
 [dependency-groups]
-docs = ["sphinx<8.2", "sphinx-book-theme>=0.3.3", "sphinx-argparse>=0.4.0", "sphinx-autodoc-typehints", "myst-nb", "sphinxcontrib-bibtex"]
+docs = ["sphinx<8.2", "sphinx-book-theme>=0.3.3", "sphinx-argparse>=0.4.0", "myst-nb", "sphinxcontrib-bibtex"]
```
ajjackson: Shouldn't there be a _minimum_ version of sphinx? Is there a specific issue with 8.2 or should this be `sphinx~=8.1` (i.e. 8.1<=v<9)?

ajjackson (commented):
Two notes. Otherwise: this is great stuff and we should maybe use it elsewhere too 😄

castep_outputs/parsers/md_geom_file_parser.py
oerc0122: Yes, sadly, Colon -> Type definition; which I have no idea how to avoid. I could do `::`. Any idea?

pyproject.toml
oerc0122: Probably sounds good!

castep_outputs/parsers/md_geom_file_parser.py
ajjackson: How about this unicode character that looks like a colon? https://www.compart.com/en/unicode/U+A789

image

ajjackson (approved):

PR #270, opened 2026-06-02
Fix reading files missing header/blank line losing first "time"

oerc0122:
When reading a file missing the header/without a blank line between header/main the first step fails to pick up time since file pointer is already on the time line.

- Fix this issue
- Add a test to ensure it doesn't happen again.
- Fix technical typing issue in `parse_castep_file`


ajjackson:
I assume this is a workaround for a very specific kind of file corruption, rather than a general behaviour issue?

The fix has a cludgy feel to it; you would think we can advance if the line is empty, rather than rewind if it isn't. But I can see how that works when we are passing around a file with some active-line state. Ultimately it is clear what happens and is well isolated to a _particular_ place in a particular file, so shouldn't be able to do much harm.

ajjackson (approved):

oerc0122:
> I assume this is a workaround for a very specific kind of file corruption, rather than a general behaviour issue?
> 
> The fix has a cludgy feel to it; you would think we can advance if the line is empty, rather than rewind if it isn't. But I can see how that works when we are passing around a file with some active-line state. Ultimately it is clear what happens and is well isolated to a _particular_ place in a particular file, so shouldn't be able to do much harm.

A behaviour issue users have supplied us with, yes. From badly chopping files up to do averaging/sampling/removing equilibration..

PR #277, opened 2026-08-03
Allow passing of function to fix_data_types

oerc0122:
- Enable expanded use of `fix_data_types`
- Add `deep` and `applydeep` functions to recurse through structures
- Use to process `lo_shifted` data.


ajjackson:
As discussed, this _could_ be used to create bewildering complexity and side-effects... but so can Classes which are already allowed. I think with a bit of discipline this can be used with well-named functions and simple lambdas or `partial(operator.foo, ...)` to do nice things.

ajjackson:
`deep`/`applydeep` seem quite fun and useful. I note that it essentially applies the same property to an arbitrary function that `to_type` applies to a class constructor. Perhaps this relationship can be made more obvious with a really ingenious name. (i.e. I can't think of one.) Also perhaps they would like to share some implementation.

oerc0122:
> `deep`/`applydeep` seem quite fun and useful. I note that it essentially applies the same property to an arbitrary function that `to_type` applies to a class constructor. Perhaps this relationship can be made more obvious with a really ingenious name. (i.e. I can't think of one.) Also perhaps they would like to share some implementation.

I did consider replacing `to_type` with a `deep` implementation, but it would require making the function only apply to `str` as part of the inner implementation which complicated things slightly and I thought if that was going to happen it could be a separate PR.

ajjackson:
Ok, let's see how this plays out. It does make things very nice at the point of use, at the cost of adding some rather clever code elsewhere. Currently the "nice point of use code" is the tested part.

Can we get some unit-testing that ensures `deep` is behaving as expected in its more generic capacity?

PR #280, opened 2026-08-06
Add flag to force output to always be in dict{file: data} format

### LINKED ISSUE #279: Always output source file
From my quick experiments, it seems that `castep_outputs` generates a dict of `<filename>: <contents>` if outputting more than one file, but just `<contents>` if outputting a single file?

This is unexpected behaviour, not least because it is not obvious whether `castep_outputs` will find more than one file.

It also makes traceability difficult for single-file output - you don't know the source filename.

I would consistently use the multi-file output.

oerc0122:
Given changing the default would be API breaking, add flag to enforce output as dictionary even for single-file processing.  

Fixes #279

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -161,12 +161,12 @@ Full usage
 
    usage: castep_outputs [-h] [-V] [-L {DEBUG,INFO,WARNING,ERROR,CRITICAL}]
                          [-o OUTPUT] [-f {json,ruamel,pyyaml,pprint,print}]
-                         [-t] [-A] [--inc-castep] [--inc-cell] [--inc-param]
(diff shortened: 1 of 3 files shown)


ajjackson:
I'm inclined to agree with the original Issue-raiser: have it always behave this way, and bump the version to 0.4.

I suppose there is an argument to make that there are effectively two modes of operation: point it at a seedname and get multiple outs, or point it at a file and get one output. But if that's the logic then it should maybe be enforced/implemented that way instead of auto-popping single-item collections.

The edge cases are pointing at a seedname or glob of filenames that happens to only match one file.

